PR #127, opened 2013-08-14
Bug fixing - wrong calculation of the calendar position

EduardTrayan:
function (this.elements.outerHeight () incorrectly calculates the calendar position and returns [object Object]

http://bugs.jquery.com/ticket/12491
http://bugs.jquery.com/ticket/12159

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -360,8 +360,8 @@
             }
             if (this.opens == 'left') {
                 this.container.css({
-                    top: this.element.offset().top + this.element.outerHeight(false),
-                    right: $(window).width() - this.element.offset().left - this.element.outerWidth(false),
+                    top: this.element.offset().top + this.element.outerHeight(true),
+                    right: $(window).width() - this.element.offset().left - this.element.outerWidth(true),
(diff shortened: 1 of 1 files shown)


longzheng:
This is now an issue with jQuery 2.0+

Using ```outerHeight()``` without a boolean will now return an object instead of a number.

It is required to have an explicit boolean ```outerHeight(true)``` to return a number.

dangrossman:
http://api.jquery.com/outerheight/ makes no mention of this.

I just tested on a page with jQuery 2.1.3 and `outerHeight()` returned a number, not an object.

longzheng:
Sorry my bad, this is a bug with jQuery UI 1.8.18 specifically https://ajax.googleapis.com/ajax/libs/jqueryui/1.8.18/jquery-ui.min.js

For example https://www.drupal.org/node/1916750

dangrossman:
Sounds like you need to not run 1.8.18 then. That's 2-3 major versions behind.

longzheng:
Yep, my bad, ignore me :(

PR #266, opened 2014-03-06
Include chosen range label as 3rd parameter to callback function

philfreo:
Important/useful for if you want to remember the user chose something like 'yesterday' and have it
always show the dynamic 'yesterday' on future visits versus a specific date.


therealgilles:
@dangrossman: I am noticing that the label is not cleared when the selected date does not correspond to the previously selected label. Is that expected?

It seems to take two selections to get the label to be 'null'.

PR #287, opened 2014-04-03
Allow start/end dates in a range to be null so there can be an 'All Time' range option

philfreo:
As originally described in #284,

Use case: I'd like to have "All Time" be a range, but want to avoid hardcoding some date like 1/1/1900. 

```
ranges: {
    'Today': [moment(), moment()],
    'This Week': [moment().startOf('week'), moment()],
    'This Month': [moment().startOf('month'), moment().endOf('month')],
    'All Time': []
}
```

The callback just returns `null` for start/end dates if this is chosen.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -24,11 +24,11 @@
                   '<div class="range_inputs">' +
                     '<div class="daterangepicker_start_input">' +
                       '<label for="daterangepicker_start"></label>' +
-                      '<input class="input-mini" type="text" name="daterangepicker_start" value="" disabled="disabled" />' +
+                      '<input class="input-mini" type="text" name="daterangepicker_start" value="" />' +
                     '</div>' +
                     '<div class="daterangepicker_end_input">' +
                       '<label for="daterangepicker_end"></label>' +
-                      '<input class="input-mini" type="text" name="daterangepicker_end" value="" disabled="disabled" />' +
+                      '<input class="input-mini" type="text" name="daterangepicker_end" value="" />' +
                     '</div>' +
                     '<button class="applyBtn" disabled="disabled"></button>&nbsp;' +
                     '<button class="cancelBtn"></button>' +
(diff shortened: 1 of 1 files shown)


mattstache:
A use case that I have run into where I don't have an end date available for my "All times" functionality. I have a page where merchants can view previous and upcoming orders. I want them to see all orders in the future. It would be great to be able to leave the end date null and have the styles updated in the calendar to show all the dates in the future as selected. Obviously there has been some discussion about this, but this is a scenario that I currently am running into.

pankajsharma120:
You can do something like - ```
ranges: {
           'All time':[null,null],
           ....
```

Edit your 

> daterangepicker.js

in line 1532 as

```
updateElement: function() {
            if (this.element.is('input') && this.autoUpdateInput) {
                var newValue = this.startDate.format(this.locale.format);
                if (!this.singleDatePicker) {
                    newValue += this.locale.separator + this.endDate.format(this.locale.format);
                }
				if(newValue == "Invalid date - Invalid date"){
					this.element.val("All Time").trigger('change');
				}
                else if (newValue !== this.element.val()) {
                    this.element.val(newValue).trigger('change');
                }
            }
        }
```
If have any doubts please let me know.

yasirrose:
> You can do something like -
> 
> ```
> ranges: {
>            'All time':[null,null],
>            ....
> ```
> 
> Edit your
> 
> > daterangepicker.js
> 
> in line 1532 as
> 
> ```
> updateElement: function() {
>             if (this.element.is('input') && this.autoUpdateInput) {
>                 var newValue = this.startDate.format(this.locale.format);
>                 if (!this.singleDatePicker) {
>                     newValue += this.locale.separator + this.endDate.format(this.locale.format);
>                 }
> 				if(newValue == "Invalid date - Invalid date"){
> 					this.element.val("All Time").trigger('change');
> 				}
>                 else if (newValue !== this.element.val()) {
>                     this.element.val(newValue).trigger('change');
>                 }
>             }
>         }
> ```
> 
> If have any doubts please let me know.

image
This is the result of this code not working for me.

PR #341, opened 2014-05-21
Month view mode & additional settings

rklachko:
Added support for month view mode (using "minViewMode" : "month")
Added new behavior settings (minViewMode, hideFormInputs, alwaysShowCalendars, autoApplyClickedRange) - All defaults were set to conserve previous behavior.
Added support for html data attributes initializing.
Minor bug fixes.


diegobfernandez:
This would be a great addition.

PR #358, opened 2014-06-24
button small class added for bs3

emamut:
Hi there,
This is a suggestion for the small buttons on bootstrap 3

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -115,7 +115,7 @@
             if (this.element.hasClass('pull-right'))
                 this.opens = 'left';
 
-            this.buttonClasses = ['btn', 'btn-small'];
+            this.buttonClasses = ['btn', 'btn-small btn-sm'];
             this.applyClass = 'btn-success';
             this.cancelClass = 'btn-default';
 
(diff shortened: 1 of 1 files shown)


karan:
Why is the default small. If I want large button, adding the BS3 `btn-lg` class does not work now.

dangrossman:
It's always been small, because there was only room for small buttons.

karan:
Maybe add the ability to override this?

So I have done this:
- `display: none` on cancel button
- larger apply button with css

There could be a much better way to do this. Maybe have a `size` option, and a `visible` option for the two buttons.

emamut:
Sounds like a great way to improve this plugin

dangrossman:
You can override it through your stylesheet (write a more specific rule than Bootstrap's to target the buttons only in the picker), or by editing your local copy. Not everything needs to be an option.

karan:
Having options make it more easily configurable and user friendly though.

dangrossman:
What classes get applied to the buttons is already an option. There doesn't need to be another "size" option that does the same thing.

PR #378, opened 2014-07-09
Allow override of "outsideClick" function

Robinson7D:
Previously triggered `hide` (which includes a submit), this allows applications to trigger `cancel`, or anything else, instead.


vikramgupta1988:
Did this PR release?

dangrossman:
no

Issue #424, opened 2014-08-18
Ability to lock range


tidemann:
Would be nice to have the ability to lock the range to i.e. 1 week or 3 days. 

dateLimit already provides a minimum range, so I'm guessing it would be possible to introduce a maximum range to fix this.

PR #427, opened 2014-08-19
Added dateLimitMin option to restrict minimum date range.  #424

tidemann:
dateLimitMin: (object) The minumum span between the selected start and end dates. Can have any property you can add to a moment object (i.e. days, months)

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -686,9 +686,9 @@
         clickPrev: function (e) {
             var cal = $(e.target).parents('.calendar');
             if (cal.hasClass('left')) {
-                this.leftCalendar.month.subtract('month', 1);
+                this.leftCalendar.month.subtract(1, 'month');
             } else {
-                this.rightCalendar.month.subtract('month', 1);
+                this.rightCalendar.month.subtract(1, 'month');
             }
             this.updateCalendars();
         },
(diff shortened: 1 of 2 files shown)


ramdog:
Hi @dangrossman - I noticed you closed an issue looking for this feature:
https://github.com/dangrossman/bootstrap-daterangepicker/issues/606#issuecomment-87392873

I have pulled in @tidemann's change from this PR and it works consistently to the existing `dateLimit` option for enforcing a max range.

sadortun:
Hi,

I would recommend the following syntax :

```
<input type="text" name="daterange" value="06/01/2015 - 06/07/2015" />
 
<script type="text/javascript">
$(function() {
        format: 'MM/DD/YYYY',
        minDate: '06/01/2015',
        maxDate: '06/30/2015',
        dateLimit: { min: { days: 6 }  , max : {days: 8 } } 
// or
        dateLimit: { [ days: 6  ,days: 8 ] } 
    });
});
</script>
````

lucasklaassen:
@dangrossman any reason this can't be merged?

dangrossman:
`this.dateLimit` gets referenced 11 times throughout the code, where this patch uses the new `this.dateLimitMin` just twice. What do you think the chances of that being bug-free are, versus there being up to 9 times the `dateLimitMin` isn't getting applied when someone will expect it to be if it's added to the list of options? Along with changing dozens of lines of code because some auto-formatter disagreed with someone else's auto-formatter, making it hard to review what actually changed, are the reason this was passed over. I don't have time to invest lots of time into reviewing and fixing partial patches that'll almost definitely create more bugs, so I pass on all but very straightforward or very thorough PRs. I also only review PRs a few times a year. It's a five year old widget that's downloaded thousands of times a week, so a couple people over the course of years interested in an option is probably below the threshold to be worth carrying around and maintaining that option forever. There's nothing wrong with adding options/features to your own copy without them being in the original branch too.

PR #489, opened 2014-10-30
Added outsideClick.daterangepicker event

kirilvit:
Added outsideClick.daterangepicker event. It's useful if we want to save or cancel editing after clicking outside and hiding dateragepicker


metaine:
how could we prevent widget to close when clicking outside the area? this is needed to scroll down the calendars on touch devices, especially when the stack on narrow screens

metaine:
that was stupid, I got to read the code. Hope commenting  .on('touchend.daterangepicker', this._outsideClickProxy) woudn't break anything

PR #500, opened 2014-11-06
Ability to show on top of input element on open

tpaksu:
This patch makes the calendar automatically show on top of the element when there's not enough space to display at the bottom of the element.


sulabhjain:
The issue which I am running into may be related to this existing one, when I scroll the page to the bottom and open date picker, it opens at the misplaced position probably top and right offset in my case are being misplaced. It works when the page scroll bar is still at the top.

github_daterangepicker

PR #508, opened 2014-11-20
Allow the picker to drop "up"

gkostov:
Added **drop** option (defaults to "down") to control whether the picker opens "down" as usual or "up" (above the picker's button)

### DIFF
--- a/daterangepicker-bs2.css
+++ b/daterangepicker-bs2.css
@@ -195,6 +195,22 @@
   content: '';
 }
 
+.daterangepicker.dropup{
+  margin-top: -5px;
+}
+.daterangepicker.dropup:before{
+  top: initial;
+  bottom:-7px;
+  border-bottom: initial;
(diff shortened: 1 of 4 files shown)


juggy:
@dangrossman Any plans on merging all those pull requests? Especially this one, it is pretty straightforward.

kirilvit:
Yeah! I need it in my project!

dangrossman:
The appearance and positioning aren't right. The picker still has an arrow at the top, which normally points to the element it's attached to, but now points to nothing. It also overlaps whatever it's attached to, covering it half with the UI's shadow.

gkostov:
Totally my fault here - I had only fixed the BS2 CSS as that is what we use. BS3 is now fixed as well and I also added to the bottom of the examples.

Cheers

dangrossman:
Thank you :)

PR #514, opened 2014-11-27
Added option to show the range view as unordered list or select box

Roywcm:
rangeView: 'ul' or 'select' 
What do you think?


Roywcm:
Any reason why this is not merged yet? Do I need to make any improvements?

PR #518, opened 2014-12-02
Added option to show/hide date range inputs

geo242:
This is useful when you only want to support using pre-defined date ranges

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -78,6 +78,10 @@
         $.each(this.buttonClasses, function (idx, val) {
             c.find('button').addClass(val);
         });
+        if (!this.showRangeInputs) {
+		    this.container.find('.daterangepicker_start_input').hide();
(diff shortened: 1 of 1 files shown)


artknight:
This is a really good idea! Should be merged asap.

artknight:
I still do not understand why this is such a problem to merge this! It does not effect anything else nor will it in the future. When users express desire for this option to be there, why are you being difficult?

dangrossman:
All pull requests against v1 were closed when v2 was released, as it was a near-complete rewrite and nothing would merge.

Please drop the attitude.

artknight:
Oh ok, i understand. Sorry about that. When was v2 released? I did a pull yesterday and I was on v1.3.2

dangrossman:
Yesterday

artknight:
you are the man!

artknight:
i do like the fact that now it opens starting with two consecutive months, not the same month twice!

PR #533, opened 2014-12-19
Allow setting options with HTML5 data attributes

pawl:
Sometimes it's necessary to set options without modifying the javascript. This pull request adds the flexibility to allow setting options through HTML5 data attributes.

This also eases the transition for people coming from bootstrap datetimepicker who already have existing code which uses data attributes: http://eonasdan.github.io/bootstrap-datetimepicker/#example5

This addresses issue #195

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,12 @@ $('input[name="daterange"]').daterangepicker(
 );
 ````
 
+Options may also be set using HTML5 data attributes. For example, timePicker="true" would be set with:
+
+```
+<input type="text" data-time-picker="true" name="daterange">
+```
+
 ## Options
 
 `startDate`: (Date object, moment object or string) The start of the initially selected date range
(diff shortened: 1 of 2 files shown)


pawl:
Any input on this?

djhopper01:
I'm not part of this project, but I tested this out because I wanted the same thing. It seems to work fine; however, this is my first day using this particular library. I'm not able to say if anything else may have been broken.

Thanks @pawl - this is a big help.

pawl:
We use these HTML5 data attributes to define bootstrap-daterangepicker's options in flask-admin because the javascript is kept in a static file that isn't using server side templating.

We could add the javascript for the datepicker to the bottom of the page template and add a bunch of IF statements for all the different options, but that would be a lot less elegant than adding the data attributes to the field.

It would be appreciated if this could get merged in so I don't have to maintain my own fork.

cavasinf:
Why JS options override HTML data options ??

As you said:
> javascript is kept in a static file that isn't using server side templating

It would be me more logic, that the HTML data attribute override the JS options.

cavasinf:
It shouldn't be: 
```JS
options = $.extend(this.element.data(), options);
```
But with `{}` at the beginning to not override ALL datapicker initalized : 
```JS
options = $.extend({}, options, this.element.data());
```

PR #553, opened 2015-01-21
Move apply/cancel buttons out of the ranges div

quinnallman:
This PR moves the apply/cancel buttons out of the ranges div so that they can still be used when ranges are hidden (e.g. when singleDatePicker = true).

### DIFF
--- a/daterangepicker-bs2.css
+++ b/daterangepicker-bs2.css
@@ -12,6 +12,10 @@
   z-index: 3000;
 }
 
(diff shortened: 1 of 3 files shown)


dangrossman:
This behavior is not desired. When only a single date picker is shown, the buttons are not necessary (clicking a date or leaving the picker express the intent without additional steps). When a single date picker and time picker are both shown, then the apply/cancel buttons are already made visible.

PR #557, opened 2015-01-26
Fix end date bigger than max date

vitorhsb:
Added a validation for the end date when selecting a custom start date,
which added to the actual difference could end up being bigger than the
max date configuratoion.


dangrossman:
You assume `this.maxDate` will be a date you can compare to, when it's not a required option. This would introduce a new bug where `endDate` could be assigned `false` instead of a date.

vitorhsb:
Ok, you're right.

With just a small edition it will be bullet proof.
if (this.maxDate && endDate.isAfter(this.maxDate)) {
    endDate = this.maxDate;
}

Added a new commit.

dangrossman:
Thanks :)

PR #562, opened 2015-01-29
Trigger change event when updating month/year

thomas-squid:


dangrossman:
Why do you want to add and trigger this event here? Nothing currently triggers a "change" event. Selecting a month/year at the top of the UI does not "change" the value of the element the picker is attached to, it only updates the calendars on screen to show dates in that month/year for selection.

thomas-squid:
I want to implement a scenario of automatically selecting the first date if the left hand calendar changes and the last date if the right hand calendar changes. If I modify this pull request would you want that functionality including?

dangrossman:
No thank you :)

PR #566, opened 2015-02-04
Add autoApply option

patmood:
Add an auto apply option that hides the apply and cancel buttons, and immediately notifies callback when calendar dates change (default is false)

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -536,6 +536,7 @@
 
         notify: function () {
             this.updateView();
+            this.updateInputText();
             this.cb(this.startDate, this.endDate, this.chosenLabel);
         },
 
@@ -734,6 +735,11 @@
                 this.hideCalendars();
                 this.hide();
                 this.element.trigger('apply.daterangepicker', this);
+
+                if (this.autoApply) {
+                    this.notify();
+                }
+
             }
         },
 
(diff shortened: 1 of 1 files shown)


bstst:
This should be merged in. As this is awesome. Why isn't it merged in?

oswaldoacauan:
:+1:

jaketoolson:
:+1: this would be great!

joseph-onsip:
:+1:

rmasters:
This would be really useful when used with `singleDatePicker`.

daterangepicker.js
```diff
@@ -290,6 +291,11 @@
                 this.timePicker12Hour = options.timePicker12Hour;
             }
 
+            if (typeof options.autoApply === 'boolean') {
+                this.autoApply = options.autoApply;
+                this.container.find('.applyBtn, .cancelBtn').addClass('hide');
```
nickolas1: <pre><code>                
if (this.autoApply) {
this.container.find(".applyBtn, .cancelBtn").addClass("hide");
                    if (this.singleDatePicker) {
                        this.container.find(".ranges").addClass("hide");
                    }
                }
</pre></code>

patmood: Agreed. I didn't consider the case when autoApply is present but set to false! I'll fix it now.

PR #581, opened 2015-02-16
Inserting actual minute/second into the list if it isn't already present

dandehavilland:
For better feedback when making finer-grained selections, in cases where `timePickerIncrement > 1` and the selected time not a round value (e.g. `11:33:03`), this will insert the actual minute and second into the time selection drop-downs.

screen shot 2015-02-16 at 18 48 40

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -1171,34 +1171,62 @@
                         selected.minute(max_minute);
                 }
 
+                var selectedMinuteDisplayed = false;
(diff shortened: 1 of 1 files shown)


dangrossman:
If you have `timePickerIncrement > 1`, I would not expect you to attach the picker to a free-form text box where people can enter times that aren't selectable. 33 should not appear in that list.

PR #584, opened 2015-02-20
Update moment.js version to 2.9.0

LukaszKrawczyk:
This update changes behaviour of `timeZone` parameter, so I think we should consider adding CHANGELOG.md file and updating package version to 1.4.0.


LukaszKrawczyk:
I've just noticed that @YidingW already did similar pull request here: dangrossman/bootstrap-daterangepicker/pull/568
The difference is that I've also updated README.md file.

PR #592, opened 2015-03-16
Fix unwanted endDate modifying maxDate.

mafredri:
Whenever endDate becomes bigger than maxDate, it is set to maxDate without cloning the object. This results in maxDate being modified every time endDate is updated.


dangrossman:
Thank you :)

PR #607, opened 2015-03-30
Call updateInputText() from setCustomDates()

joseph-onsip:
This keeps the input element synchronized when the user clicks a date on a calendar, but doesn't click the apply button. Related to #332

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -779,6 +779,7 @@
 
             this.updateView();
             this.updateCalendars();
+            this.updateInputText();
         },
 
         clickDate: function (e) {


dangrossman:
The input should not be synchronized with discarded changes.

joseph-onsip:
Oh, perhaps I made this change in the wrong place. The use case I have in mind is a form whose submit button uses the picker's `startDate` and `endDate`. If the user opens the picker, clicks a date range, then clicks the submit button, the `startDate` and `endDate` values correspond to the dates they clicked (as I would have expected), but the input element does not. Would it be preferable to call `updateInputText` [here](https://github.com/dangrossman/bootstrap-daterangepicker/blob/c44faec07098c63fdd129cf4315849f4c1e37e0b/daterangepicker.js#L822) instead?

PR #611, opened 2015-04-01
Require only core dropdown bootstrap component

Braunson:
Included the local Twitter Bootstrap core component (dropdown.js) that is used instead of the extra overhead for the whole library.

Additionally, people can use it in other libraries, i.e. Foundation.

I will be working on a fully integrated foundation version sooner than later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ The component can also be used as a single date picker by setting the `singleDat
 
 ## Usage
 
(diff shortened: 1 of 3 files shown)


dangrossman:
This repository does not include the Bootstrap library, so there is no reduction by adding part of it. End-users are free to download only the dropdown script or use their favorite package/dependency manager.

PR #624, opened 2015-04-15
Added prevIconClasses and nextIconClasses options

apancutt:
Added ability to override/customise the classes applied to the 'previous' and 'next' buttons.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,10 @@ $('input[name="daterange"]').daterangepicker(
 
 `cancelClass`: (string) CSS class string that will be added to the cancel button
 
+`prevIconClasses`: (array) CSS class names that will be added to the previous button icon
+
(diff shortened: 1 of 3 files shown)


dangrossman:
Why does this need to be an option, rather than editing your copy of daterangepicker.js? Are you creating multiple pickers on a page with different arrow styles in each one?

apancutt:
Want it as an option so I don't have to modify my local file as I'd then
have to manage my own version. This option could be useful for others too.

In my case its required because I'm using the Glyphicons Pro pack instead
of limited Bootstrap bundle, and therefore icon names are different.

dangrossman:
I think too many options for every little bit of the HTML makes the thing harder to use, understand and maintain for others. There's nothing wrong with having your own copy. It's not like an OS kernel module, it's just a calendar widget for a webpage; if you're happy with it, you don't have to ever update it.

PR #640, opened 2015-04-30
isInvalidDay - users can set individual days as being invalid

gmarotta:
Analogous to jquery-ui's onBeforeDay option, this lets users set a method to determine if individual days are invalid (weekends, holidays, etc).  For example

```javascript
   isInvalidDay: function(day) {
      if (day.day() === 0 || day.day() === 6) {
         return false;
      } else {
         return true;
      }
   }
});
```

will block out all Sunday and Saturdays

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -282,6 +282,11 @@
                 this.timePicker12Hour = options.timePicker12Hour;
             }
 
+            if (typeof options.isInvalidDay === 'function') {
+                this.isInvalidDay = options.isInvalidDay;
+            }
+
+
             // update day names order to firstDay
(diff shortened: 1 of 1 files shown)


digitalvapor:
:+1: confirmed that this works!! super helpful.

digitalvapor:
@dangrossman what is criteria for items like this to be included?

dangrossman:
This probably *will* be included when I get a chance. I closed all open PRs against v1 because v2 is out, and was a near-complete rewrite -- nothing would have merged, and many of the bugs were fixed as a side effect.

digitalvapor:
@dangrossman thanks for taking time to explain :+1:

PR #645, opened 2015-05-08
Problem: sometimes it is needed to react right after the user selects…

mindreframer:
… a date, before he clicks on apply button

Solution: 
- trigger the ‘clickDate.daterangepicker’ event in clickDate function
- update Readme to document that behaviour


mindreframer:
closing it, will make a pull request from the company's github account.

PR #646, opened 2015-05-08
Problem: sometimes it is needed to react right after the user selects…

mindreframer:
… a date, before he clicks on apply button

Solution: 
- trigger the ‘clickDate.daterangepicker’ event in clickDate function
- update Readme to document that behaviour


dangrossman:
I don't think this is needed enough to add another event and more documentation. Either maintain your own copy, or add the click handler for those table cells outside of the DRP code.

Luk4:
I also added this event to my copy, very handy to update a label as the user clicks 😀

Issue #652, opened 2015-05-13
From and to field values get overwritten on hover


andri:
I'm using the range picker in a product and noticed a small issue with the from and to fields.

If a person changes the dates and then moves the mouse over the range picker (Today, yesterday etc.) without hitting enter or tabbing away from the field, clicking apply will not use the updated date information but the dates last hovered over with the mouse.

To reproduce:

* Change from or to with the keyboard.
* Now without hitting enter or tabbing from the field move the mouse over the range picker.
* Now click "Apply".

![image](https://cloud.githubusercontent.com/assets/123914/7610688/428f00c6-f96e-11e4-802c-167acb6ae8be.png)

andri:
This is probably more of an issue in my case since I have altered the design and moved the apply button but it can very well happen using the default style.

dangrossman:
It is. And because you want it to work differently than the default, you'll have to edit your own copy to resolve this. I recommend removing the code that alters the dates on hover.

andri:
Thanks for taking the time.

Its more of an usability issue, it also happens in the live example but because the apply button is below the form users are not affected as often.

dangrossman:
Originally those boxes were not even editable. They were meant to reflect the currently selected dates, or what dates would be selected if you clicked on a range label while hovering over them, to make that more clear before selecting. Making them editable was suggested and added by someone with a pull request at some point.

andri:
@dangrossman, I made a pull request. Using different events for monitoring the fields does the trick pretty well. Check it out.

PR #653, opened 2015-05-13
Switched to using the input and propertychange events to update the calendar

andri:
This makes the calendar more responsive and fixes an issue where changes get lost when the user does not use the enter key before hitting apply in some situations. 

See issue #652 for more details.


andri:
I realised I should have added these new events and not removed the change event. Will fix that. There are also some IE8-9 issues that need a better look.

PR #657, opened 2015-05-16
Allows singleDatePicker to appear left, right or centered

gutschik:
This allows the picker to be aligned on a direction other than the default right


motokimizoguchi:
+1

PR #658, opened 2015-05-19
Localization options

unrealsolver:
I've just implemented better localization options for the widget.
Now `options.locale` accepts:
* `Object : [locale object]` - old approach to set locale
* `String : 'moment'` - use locale described in `moment.locale()`. Locales for momentjs should be included to your app separately, or you can use complete momentjs bundle with all locales.
* `String : like 'en', 'ru', 'ja', etc` - use specified locale. I've added only locales for russian (my native lang.) and japanese (translated by google). Also, this approach does not affect momentjs, so names of the days/months will not be translated.
calendar-localization

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ $('input[name="daterange"]').daterangepicker(
 
 `separator`: (string) Separator string to display between the start and end date when populating a text input the picker is attached to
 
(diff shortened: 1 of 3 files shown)


7dir:
add to docs

PR #659, opened 2015-05-20
Keep label consistent and change the default value of timePickerIncrement.

kangkona:


dangrossman:
"Week" is not consistent with anything else; it's too wide to be a column label in the calendar.

Changing a default means potentially breaking websites for anyone doing automatic deployments and using the default setting. It is not to be changed so lightly.

PR #660, opened 2015-05-22
Apply dateLimit when input is typed in

andersstorhaug:
The `dateLimit` works well when dates are selected via clicks -- this is just a simple fix to apply `dateLimit` when inputs are typed in as well.


dangrossman:
Good idea. However, rather than duplicate this check in multiple places, it should probably be moved to `setCustomDates` which is called by `clickDate` and `inputsChanged`.

andersstorhaug:
Thanks for pointing that out -- I'll look into doing that and add another commit.

PR #689, opened 2015-06-30
add second select option

billdc:
add two options , showsecond secondIncrement,
use these two options to setup seconds.


aralroca:
This was useful.... Why is closed? Seconds and milliseconds precision is good...

dangrossman:
2 people in 5 years is not enough for me to maintain the code for your use case forever.

Issue #225, opened 2014-01-20
When calendars are stacked, "to" calendar is above "from" calendar


nonplus:
When you use the daterangepicker on a narrow page, the calendars are stacked on top of each other.  However the "starting" calendar comes **after** the "ending" calendar.

![image](https://f.cloud.github.com/assets/1023803/1958771/afcfc9d0-821f-11e3-97c9-075748d7c753.png)

FWIW, when you show the calenders with the "right" option, they are stacked correctly.

mhenley-uk:
+1

cnizzdotcom:
Sorry but this bug makes an otherwise wonderful plugin useless.

vodich:
@robinfai +2 :)

laurent1:
`.daterangepicker {
    min-width: 670px;
}`

dangrossman:
What's that supposed to mean, @laurent1? This bug was already fixed, the calendars stack in the correct order now.

laurent1:
@dangrossman I am using the version on npm, which points to version 1.3.10. So, I am a could of years behind. Thanks for the quick response.

dangrossman:
This library is also published on npm, under https://www.npmjs.com/package/bootstrap-daterangepicker

indra1:
I use daterangpicker 2.1.5 and I have this issue on one row.

![image](https://user-images.githubusercontent.com/1842674/31939247-71e55508-b8c2-11e7-91ad-330def6d0200.png)


This is how I setup the date rang picker

```
function activateDateRagePicker(field) {

        var start = moment().startOf('month');
        var end = moment().endOf('month');

        function cb(start, end) {
            field.html(start.format('DD/MM/YYYY') + ' - ' + end.format('DD/MM/YYYY'));
        }

        field.daterangepicker({
            locale: {
                format: 'DD/MM/YYYY'
            },
            startDate: start,
            endDate: end,
            ranges: {
                'Today': [moment(), moment()],
                'Yesterday': [moment().subtract(1, 'days'), moment().subtract(1, 'days')],
                'This Week': [moment().startOf('week'), moment().endOf('week')],
                'This Month': [moment().startOf('month'), moment().endOf('month')],
                'This Year': [moment().startOf('year'), moment().endOf('year')],
            },
        }, cb).on('hide.daterangepicker', function (e) {
            $(e.target).val();
        }).on('cancel.daterangepicker', function (e) {
            $(e.target).val('')
        });
[...]

PR #690, opened 2015-06-30
[FIX] calendars are stacked inverted #225

sylvainbannier:
This is an attempt to fix issue #225.
This fix is tested on Chrome 44 and IOS7.

Also tested on Firefox 38 but Firefox don't stack calendars at all.

A side effect is that predefined ranges move to the right as shown in the following screenshot.


dangrossman:
I'm not going to take a PR that degrades the layout on desktop to fix the layout on mobile. They are in that order [predefined ranges list that may expand to the right to choose custom dates, calendar, calendar] when attached to an element on the left side of the screen for a reason.

PR #696, opened 2015-07-05
Fix browserify/commonjs bug : plugin object isn’t exported

harobed:
Fix browserify/commonjs bug : plugin object isn’t exported

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -10,10 +10,10 @@
 
   if (typeof define === 'function' && define.amd) {
     define(['moment', 'jquery', 'exports'], function(momentjs, $, exports) {
-      root.daterangepicker = factory(root, exports, momentjs, $);
+      root.daterangepicker = factory(root, momentjs, $);
     });
 
-  } else if (typeof exports !== 'undefined') {
+  } else if(typeof module === "object" && typeof module.exports === "object") {
     var momentjs = require('moment');
     var jQuery;
     try {
@@ -23,14 +23,13 @@
       if (!jQuery) throw new Error('jQuery dependency not found');
     }
 
-    factory(root, exports, momentjs, jQuery);
-
+    module.exports = factory(root, momentjs, jQuery);
   // Finally, as a browser global.
   } else {
-    root.daterangepicker = factory(root, {}, root.moment || moment, (root.jQuery || root.Zepto || root.ender || root.$));
+    root.daterangepicker = factory(root, root.moment || moment, (root.jQuery || root.Zepto || root.ender || root.$));
   }
 
-}(this, function(root, daterangepicker, moment, $) {
+}(this, function(root, moment, $) {
 
     var DateRangePicker = function (element, options, cb) {
 
@@ -1336,7 +1335,7 @@
 
     };
 
-    $.fn.daterangepicker = function (options, cb) {
+    return $.fn.daterangepicker = function (options, cb) {
         this.each(function () {
             var el = $(this);
             if (el.data('daterangepicker'))
@@ -1345,5 +1344,4 @@
         });
         return this;
     };
-
 }));


dangrossman:
This will break other integrations with other package systems / front-end frameworks. I don't use them all, so I can't say which, I just know some of the code you deleted was *added* specifically for em.

daterangepicker.js
```diff
@@ -23,14 +23,13 @@
...
-    factory(root, exports, momentjs, jQuery);
-
+    module.exports = factory(root, momentjs, jQuery);
   // Finally, as a browser global.
   } else {
-    root.daterangepicker = factory(root, {}, root.moment || moment, (root.jQuery || root.Zepto || root.ender || root.$));
+    root.daterangepicker = factory(root, root.moment, (root.jQuery || root.Zepto || root.ender || root.$));
   }
 
-}(this, function(root, daterangepicker, moment, $) {
```
harobed: I can remove daterangepicker if this function because it is never used.

daterangepicker.js
```diff
@@ -10,10 +10,10 @@
 
   if (typeof define === 'function' && define.amd) {
     define(['moment', 'jquery', 'exports'], function(momentjs, $, exports) {
-      root.daterangepicker = factory(root, exports, momentjs, $);
```
harobed: I can remove exports because I remove daterangepicker argument in factory function.

daterangepicker.js
```diff
@@ -10,10 +10,10 @@
...
+      root.daterangepicker = factory(root, momentjs, $);
     });
 
-  } else if (typeof exports !== 'undefined') {
+  } else if(typeof module === "object" && typeof module.exports === "object") {
```
harobed: This break nothing.

daterangepicker.js
```diff
@@ -23,14 +23,13 @@
       if (!jQuery) throw new Error('jQuery dependency not found');
     }
 
-    factory(root, exports, momentjs, jQuery);
-
+    module.exports = factory(root, momentjs, jQuery);
```
harobed: This break nothing.

daterangepicker.js
```diff
@@ -1336,7 +1335,7 @@
 
     };
 
-    $.fn.daterangepicker = function (options, cb) {
+    return $.fn.daterangepicker = function (options, cb) {
```
harobed: This break nothing.

harobed:
@dangrossman I've appended some comment in my patch, I think this patch break nothing.

daterangepicker.js
```diff
@@ -23,14 +23,13 @@
       if (!jQuery) throw new Error('jQuery dependency not found');
     }
 
-    factory(root, exports, momentjs, jQuery);
-
+    module.exports = factory(root, momentjs, jQuery);
   // Finally, as a browser global.
   } else {
-    root.daterangepicker = factory(root, {}, root.moment || moment, (root.jQuery || root.Zepto || root.ender || root.$));
+    root.daterangepicker = factory(root, root.moment, (root.jQuery || root.Zepto || root.ender || root.$));
```
dangrossman: Here you change the function signature and remove `moment`, which is needed by Meteor.js users as the `moment` object is not attached to the window (root) for them. This change would mean they have to go back to maintaining a fork instead of using this package.

harobed: Yes, this is a mistake, I've worked on old version. I'll fix that.

harobed:
I've fixed : 

```
    root.daterangepicker = factory(root, root.moment, (root.jQuery || root.Zepto || root.ender || root.$));
```

to 

```
    root.daterangepicker = factory(root, root.moment || moment, (root.jQuery || root.Zepto || root.ender || root.$));
```

Now I think that all is clean.

PR #697, opened 2015-07-08
Add option to hide range panel

zation:
`showRange`: (Boolean) Set to false to hide range panel, and click date directly to apply and update input text. Default value is true.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ node_modules
 npm-debug.log
 *.sublime-workspace
 *.sublime-project
+.idea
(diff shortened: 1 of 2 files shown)


artknight:
Why is this not merged yet?

dangrossman:
(1) Because it only duplicates functionality that already exists: this doesn't appear to do any more than the `autoApply` option that already hides the apply/cancel buttons and applies the selection as soon as you choose a date

(2) Because I have other obligations and get to pull requests when I get to them

(3) Because I don't accept every pull request; I am the steward of this code, who will have to maintain it for years to come, every PR making it more complex to do so, while those who submit the PRs have no such concerns

artknight:
This is a really small change and "showApply" does not address the input fields. Even though there is way to hide the fields via CSS, but sometimes this is a much nicer way. Anyways, this is your show.

PR #701, opened 2015-07-10
new behavior on range click

borodulin:
Thank you for the wonderful widget, but I am constantly confronted with the need to change behavior when clicking on the range. Could you add a property, so as not to close the calendar when you select a specified range?

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -132,6 +132,7 @@
             this.autoApply = false;
             this.singleDatePicker = false;
             this.ranges = {};
+            this.rangeAutoClose = true;
 
             this.opens = 'right';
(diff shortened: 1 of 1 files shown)


borodulin:
http://jsfiddle.net/borodulin/vsx4kn7k/4/

dangrossman:
Why?

borodulin:
Users often do not understand what is happening. They need to clearly see the period and be able to clarify it. In the current implementation, when the dialog is closed, they open it again to make sure what period selected.

vperron:
We are facing the issue too and I agree with @borodulin that making this an option if better than the hack we have implemented here (when it closes, open it again :/ )

PR #715, opened 2015-07-17
allow fireCallback to be called externally

719media:
I noticed there was no way to "externally" fire the callback. So when I had a custom callback function, I could not call it directly from javascript. I wanted to be able to do:

$('#element').data('daterangepicker').fireCallback();

I'm open to other solutions, but basically the problem was such that when I had a custom callback, it was not being fired on initial "setup" of the plugin, and also not fired on subsequent method calls like setStartDate since they were not going through the normal plugin process of render calendar/update selection/hide calendar.

Also, minor spacing cleanup.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -20,7 +20,7 @@
       jQuery = require('jquery');
     } catch (err) {
       jQuery = window.jQuery;
-      if (!jQuery) throw new Error('jQuery dependnecy not found');
+      if (!jQuery) throw new Error('jQuery dependency not found');
     }
 
     factory(root, exports, momentjs, jQuery);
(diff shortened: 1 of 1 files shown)


dangrossman:
> So when I had a custom callback function, I could not call it directly from javascript.

Sure you can.

    function myCallback(start, end) { alert("Do something"); }

    var start = moment(), end = moment.add(1, 'day');
    $('#element').daterangepicker({ startDate; start, endDate: end }, myCallback);
    
    //fire on initial setup
    myCallback(start, end);
    
    //fire when changing dates
    $('#element').data('daterangepicker').setStartDate(start);
    $('#element').data('daterangepicker').setEndDate(end);
    myCallback(start, end);

719media:
The difference is that you are tracking start/end values outside of daterangepicker. I understand there are methods around this, but they all require additional variables outside of the scope of the daterangepicker.

Say, for example, you are using data-start-date and data-end-date attributes in html... or maybe you're just allowing daterangepicker to use its own default (today). You have no way of retrieving the start/end values externally.

$('#element').daterangepicker({}, function(start, end, label) {
console.log(start);
});

//what is the startdate of the daterangepicker? not directly accessible.

an alternative would be a getter for start_date/end_date, perhaps.

dangrossman:
You can use `$('#element').data('daterangepicker').startDate` and `$('#element').data('daterangepicker').startDate`

719media:
OK, sounds like a good alternative. Thanks for the chat.

PR #723, opened 2015-07-21
Remove unnecessary engine version restrictions in package.json

SpaceK33z:
These engines aren't used anywhere anyway and give a confusing warning when `npm install`-ing.


dangrossman:
Thanks. I didn't create the file so didn't know if there was a reason for that to be there.

PR #725, opened 2015-07-22
Allow to leave blank the field in init plugin

sb4m:
Allow to leave the input empty


dangrossman:
I'm open to thinking about an option for this, but it won't be implemented like this: `leaveBlank` is vague and you actually have the element filled in if `leaveBlank` is true rather than if it's false.

sb4m:
Yes my pull sucks, I rollback to an older version

PR #734, opened 2015-07-27
when used in ember app a _super function gets added to classes string

serega104:
ember_bug


dangrossman:
Thank you for the fix

PR #743, opened 2015-08-01
Use jQuery if already loaded

paltman:


dangrossman:
I can only hope this is a good PR as I'm not equipped to test it.

PR #752, opened 2015-08-06
add timePickerHourLimit config option

Zane-XY:
add timePickerHourLimit configure option, so you can use it like:

```javascript
timePickerHourLimit:{left:[9, 14], right:[14, 18]},
```


dangrossman:
Thanks for the PR, but I'm passing on this.

coelhoricardo:
I would appreciate this feature.

PR #753, opened 2015-08-06
Automatic 24 hours detection

xibman:
You can use moment locale to detect time format


dangrossman:
A "dynamic default" is going to surprise users of this library when it works differently for some of their users than others. If they want to initialize that option using the locale, they can always do so themselves, aware of the consequences.

PR #757, opened 2015-08-07
add before show event

bobmajdakjr:
allows you to completely reconfig the widget right before showing if needed. our form is two fields, and i needed to be able to change settings of each when activated. an example resetting the maxDate prior:

InputStart.on('before.daterangepicker',function(e,picker){ picker.maxDate = moment(InputEnd.val(),picker.locale.format); });

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -999,7 +999,10 @@
 
         show: function(e) {
             if (this.isShowing) return;
-
+            
+            // Allow for reconfigure prior to display.
+            this.element.trigger('before.daterangepicker', this);
+            
             // Create a click proxy that is private to this instance of datepicker, for unbinding
             this._outsideClickProxy = $.proxy(function(e) { this.outsideClick(e); }, this);
             // Bind global datepicker mousedown for hiding and


dangrossman:
Thank you for offering this PR, but I'm going to decline it. I don't think this is a common use case.

bobmajdakjr:
its a 1 line feature that does nothing but add value. i recommend it, else i'll maintain my fork which may fragment users. i'll even write the 1 line of documentation for you.

dangrossman:
I recommend binding to the click and focus event of your input, which triggers this very same event without requiring any new code or documentation for everyone else.

bobmajdakjr:
my original attempt went that direction, but that seems to be an async race condition. sometimes it would appear to work, sometimes it would not. i wrote it off as my focus() sometimes going before your focus() and other times visa versa.

dangrossman:
You can add a delay to your function to avoid that

dangrossman:
You can do a couple things: call `show()` again on the picker to make sure your changes take effect before re-running the code, attach a new `.daterangepicker()` to the input with your new options and call `.show()` on that...

bobmajdakjr:
I love how the solution to avoid one line of elegance is convolution. computers and libraries in general are suppose to make life easier. I appreciate your work up to this point, but I think to maintain my fork then for now.

dangrossman:
Just trying to help, since you seemed hesitant to alter your own copy for some reason.

PR #759, opened 2015-08-08
Feature/is invalid date

bottomline:
Recreating this PR against v2 of daterangepicker.

Analogous to jquery-ui's onBeforeDay option, this lets users set a method to determine if individual days are invalid (weekends, holidays, etc). For example

   isInvalidDate: function(day) {
      if (day.day() === 0 || day.day() === 6) {
         return false;
      } else {
         return true;
      }
   }
});
will block out all Sunday and Saturdays

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -248,8 +248,8 @@
         if (typeof options.linkedCalendars === 'boolean')
             this.linkedCalendars = options.linkedCalendars;
 
-        if (typeof options.isInvalidDay === 'function')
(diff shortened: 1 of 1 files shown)


dangrossman:
Thank you for this. I'm away on vacation right now and have irregular access to a computer. I'm going to take this PR when I have a chance to try it and add it to the documentation. It's no use if nobody knows it exists :)

bottomline:
Swap the false / true in my above example -- I was tired when I wrote it.

PR #766, opened 2015-08-11
fix opens right so the dropdown-menu is actually aligned to the right

daniel-wer:
Setting opens to 'right' while creating the daterangepicker did not have the intended effect as the daterangepicker was aligned to the left nevertheless. This can be seen using the Configuration Generator on your website as well.

open-right-bug

This pull request fixes this behaviour and the daterangepicker will be aligned to the right es expected when setting opens to 'right'.

image

### DIFF
--- a/daterangepicker.css
+++ b/daterangepicker.css
@@ -75,7 +75,7 @@
 .daterangepicker.opensright:before {
   position: absolute;
   top: -7px;
-  left: 9px;
+  right: 9px;
   display: inline-block;
   border-right: 7px solid transparent;
   border-bottom: 7px solid #ccc;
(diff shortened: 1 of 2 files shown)


dangrossman:
Your screenshot shows the intended behavior.

"Opens: right" means it's anchored to the left side of the input and expands in width to the right when the calendars need to be displayed. It looks a little strange since you wouldn't "opens: right" on something at the right edge of the page where there isn't room to open to the right.

Your second screenshot shows the intended effect of "opens: left" setting.

This isn't a bug.

daniel-wer:
Sry for the misinterpretation of the opens setting, but nevertheless, in that case opens left is not working as intended:

image

I just found out that it does indeed work if you explicitly set it to 'left', which the Configuration Generator is not doing when choosing 'left'. So it seems to be a problem with the default case handling.

dangrossman:
Fixed, thanks

PR #772, opened 2015-08-15
Support for function call in data-ranges

l3ehnam:
setting moment() in data-api e.g. <input type="text" data-ranges='{"Today": {"start":"moment()"}}'>


dangrossman:
It's cool, but everyone seems to be getting by without that, so I'm going to pass on including an undocumented and unexpected `eval()` in a UI widget.

PR #774, opened 2015-08-18
A less error-prone way of array iteration

Reggino:
Thanks for this!

Please consider this way of iterating arrays to prevent issues with e.g. a modified ```Array.prototype```. More info can be found here:  https://github.com/skratchdot/react-bootstrap-daterangepicker/pull/28


dangrossman:
You probably shouldn't add properties to every array, generally... I think the assumption that an array I just created only has the elements I just added to it is a sane one.

PR #777, opened 2015-08-20
Added options to disable date ranges, rather than removing them

mtraynham:
This adds the ability to optionally disable the ranges in the list rather than removing them.  You can also pass a predefined `disabledRanges` array of strings to disable ones that **do** meet the range criteria.

### DIFF
--- a/daterangepicker.css
+++ b/daterangepicker.css
@@ -339,3 +339,10 @@
   border: 1px solid #08c;
   color: #fff;
 }
+
+.daterangepicker .ranges li.disabled, .daterangepicker .ranges li.disabled:hover {
+  background: #f5f5f5;
+  border: 1px solid #f5f5f5;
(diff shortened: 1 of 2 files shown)


dangrossman:
What is the reason for defining ranges that can't be used?

mtraynham:
Call it *teasing* functionality, if you will.  I wanted to be able to show that functionality is supported at given time intervals, but data/user permissions are not there to properly handle that functionality.  When the date ranges are hidden, it's unknown the user, that they could potentially view a larger/smaller increment.  When it's disabled, they understand it's just not available.

dangrossman:
I can see how you might use that, but it doesn't seem common/demanded enough to warrant yet another option and my perpetual support of the code. I'm passing on this PR, but thanks for offering it.

PR #786, opened 2015-08-28
add return value to support amd and commonjs

songquanwang:
Dear:
          In the line 11-31 of  this src file ,your export this module ,but the factory function does not has a return value.I think it should return an object. thanks!

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -1455,4 +1455,5 @@
         return this;
     };
 
+	return DateRangePicker;
 }));


dangrossman:
I have no way to evaluate the correctness or usefulness of this PR. I don't use amd or commonjs.

songquanwang:
your can ref the backbone.js https://github.com/jashkenas/backbone/blob/master/backbone.js

PR #787, opened 2015-08-28
Enhancement make it possible to determine if ranges appear on the right or left side of calendars

pedrocatre:
The changes add a new option named rangesPositionInRelationToCalendars that allows the user to determine the position of the ranges section in relation to the calendars. The position can be *'after'* (default option) or *'before'* the calendars

This change was introduced because when the date range selector is on the left it is confusing for users to open the select range, see the default ranges, click on custom range and have all the ranges be pushed to the right by the calendars that appear. This is only a problem when the date range picker is on the left of the application and it opens to the right and it has ranges.

## Current behaviour:

* User clicks on the input for the date range picker to open: 
screen shot 2015-08-28 at 12 23 41

* User clicks on custom (the calendar pushes the ranges to the left):
current_behaviour

## New behaviour
Value for rangesPositionInRelationToCalendars is configured to *'before'*:

* Again user clicks on the input for the date range picker to open: 
screen shot 2015-08-28 at 12 23 41

* User clicks on custom. The calendars appear on the right of the range. No content is pushed:
new_behaviour

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -64,6 +64,8 @@
         if (this.element.hasClass('dropup'))
             this.drops = 'up';
 
+        this.rangesPositionInRelationToCalendars = 'after';
+
         this.buttonClasses = 'btn btn-sm';
         this.applyClass = 'btn-success';
         this.cancelClass = 'btn-default';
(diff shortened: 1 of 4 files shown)


dangrossman:
Improving this was on my TODO list too. I'm curious about one thing. Why add a new option, rather than swapping the order based on the already existing "opens" option?

pedrocatre:
I thought about using the "opens" option. Personally, for my use case, it does not matter. However, it occurred to me that some developers might like to be able to control this.

One case that came to mind where being able to control this might be nice was if:
* the date range picker input is on the left
* there are no predefined ranges
* the autoApply option is set to false

In that case the calendars immediately open up and I believe it would be better to have the "Apply" button on the right because it looks better to me and because the normal flow of actions is to 1st pick the starting date, then the end date and then click apply, and by having the "Apply" button on the right this flow is optimised because the "Apply" would be closer to the end date calendar (assuming in most cases the user chooses the end date on the 2nd calendar and not the 1st).

most_common_user_flow

I can change this to use the "opens" option if you would like. Even if there is a loss of flexibility it still could make sense given that the API would be less complex by having one less option.

pedrocatre:
Let me know if you want me to update this pull request to use the "opens" option, if you like it like this given the reasoning I presented above, or if you intend to implement this yourself later in a different way.

dangrossman:
I decided to go with the simple route, which is expanding the widget in the direction of the `opens` setting, rather than adding a whole new option. It's only a couple lines of code:

https://github.com/dangrossman/bootstrap-daterangepicker/commit/77169c631b3cda69e4ce6833a6497a372b9e44b4

Issue #793, opened 2015-09-02
Empty date


ehhc:
It would be nice to have the possibility to not select a date at all. Especially if a single date picker is used there are different use cases where it would be useful to give the user the possibility not to select a date (e.g. if the date represents the date of an event that might not happened yet).

chillyistkult:
Yes that would be really nice. The "Clear" button hack described in the docs isn't a good solution.

superdummy:
+1: Just searched exactly for this feature. Would be really great to have this feature. There are alot of use cases where this makes sense. Thanks very much!

ehhc:
Yes, as i said, i think this would be a nice and usefull feature too.
As you see i've commited a fix for this problem and i hope @dangrossman  will accept my pull request and merge my changes (or an other solution) into the official code base.
Until then you're welcome to use my fork.

pascalmartin:
+1: I also need this functionality

legrego:
+1: this would be great

dangrossman:
This date picker appears to support no date selected. This one seems a better fit for your needs:

http://eternicode.github.io/bootstrap-datepicker

chillyistkult:
@dangrossman it's not really helpful to just point to another repository. There are dozens of other datepickers that support this feature, but we bring it up here for a reason.

superdummy:
Hello @dangrossman

I spend considerably amount of time trying to find and evaluate all (bootstrap/jqurey compatible) date and time(!) pickers and ended up with yours.
I am not Gartner, but you are in the magic quadrant on the top right corner! Leader in functionality and(!) UX. I know the other pickers but they all had major lack in functionality (time picker) or a bad UX. And take this as a big kudos! Your coding and UX skills are the best! 

Everone who +1 and also the first poster who even wrote a patch, showed interest in making your cool plugin even more cooler. And the Number of +1 in such a short amount of time shows how important this feature is. From an UX standpoint for a Single Date Picker this is one of the most fundamental UX-Patterns, as empty means, you as a user,  have to do something...  

Please think about it, it will make your plugin even more better and the world too.

alexdd55:
:+1:

uniteddevelopers:
@alexdd55 and all that are coming to this ticket: Very good news. DateRangePicker now supports empty Dates. See this ticket and the answers from dangrossmann: 
https://github.com/dangrossman/bootstrap-daterangepicker/issues/815

ehhc:
[...]

PR #794, opened 2015-09-02
Issue #793 - Empty date

ehhc:


dangrossman:
This is a hack that doesn't cover many combinations of settings. It will lead to bugs, confusion, and lots of new issue reports. It is not so easy to support having no date selected in this codebase.

marceloudi:
@dangrossman can you explain "too many combinations" with an real example?

In my case (and other users too):
- 90% of inputs in a huge software _must_ have initial empty values, so I need to _rewrite_ code to achieve this?
- Some filter fields _can_ be filled at start because of server stored "User Preferences". Ex: "Last 30 days", but in other cases the user clear the value

So, my question is: _rewriting_ code using "autoUpdateInput + callback" is _better practice_ than provide an initial option?

dangrossman:
Your question: Maybe not, but nobody has coded that setting, and it would require more code than this PR contained to work with all the other settings this library has. I can't give specific examples, sorry. Out at dinner, and  my old open source repos aren't first priority when I'm working either.

marceloudi:
Ok Dan! I understand it. I am a open source collaborator at ubuntu, so we share almost the same situation.
So I've tested @ehhc solution to exhaustion, and did not found any breaks.

Maybe you can considerate this proposal as a good help to you efforts.

I sincelery understand you.

Regards.

PR #806, opened 2015-09-16
Declare `html` var. 

voidstardb:
This undeclared variable causes an error if this file is included in strict environments. Also, the undeclared variable was leaking data into the global context.


dangrossman:
Thanks, unintentional!

PR #812, opened 2015-09-23
No callback on empty options startDate & endDate

asborisov:
Changes to fix luisfarzati/ng-bs-daterangepicker#59
When no startDate & endDate in init options and just clicking OK button no callback executing

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -39,7 +39,9 @@
         this.parentEl = 'body';
         this.element = $(element);
         this.startDate = moment().startOf('day');
+        this.startDateIsNull = false;
         this.endDate = moment().endOf('day');
+        this.endDateIsNull = false;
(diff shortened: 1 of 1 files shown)


dangrossman:
I don't know what this is doing or why. If you don't set a `startDate` and `endDate` options, that just means the default value of today's date gets used. They're never null internally. A callback is never fired with null dates.

asborisov:
Yes, you are right. 
But you directive code in luisfarzati/ng-bs-daterangepicker#59 you will see that the only way it can react on value change in daterangepicker is callback function. 
Now lets imagine situation when there is no default value in options and we want to set range for today. We click OK button, but callback function is not executing and we get nothing in ngModel value.
So may code only set flag that there is no startDate or endDate in options and when we click OK button first time execute callback, because it's the only way we can get new value in directive.

PR #816, opened 2015-09-28
check window exists before binding

peter-mouland:
To ensure plugin doesn't break server-side rendering


peter-mouland:
updated and added comments as to why we might do something weird looking for 'client-side' js.

PR #821, opened 2015-10-02
Adding rangesRows option

sezeresen:
daterangepicker

### DIFF
--- a/daterangepicker.css
+++ b/daterangepicker.css
@@ -322,6 +322,15 @@
   padding: 0;
   width: 160px;
 }
+.daterangepicker .ranges ul.multi {
+    float: left;
+    margin-left: 5px;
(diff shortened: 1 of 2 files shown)


sezeresen:
Option name can be `rangesPerColumn` also.

peter-mouland:
`rangesPerColumn` sounds good, default to `1`.

but this code is a little lazy though. a new ul for each column? not close to being semantic.  better would be to set widths for the li's based on the number of columns

dangrossman:
I am quite happy to review my PRs myself, peter.

dangrossman:
I'm going to opt not to take and support this feature. There's been no demand for something like this before, and multi-column lists can be accomplished purely in a stylesheet without library changes.

Issue #825, opened 2015-10-08
Rendering Issue in Firefox


rksaxena:
<img width="764" alt="screen shot 2015-10-08 at 2 29 35 pm" src="https://cloud.githubusercontent.com/assets/5802062/10361986/365b9416-6dc9-11e5-8270-277585f66f35.png">

I am using this date-range picker with following configuration:
```javascript
    $(that).daterangepicker({ 
        autoApply: true, 
        locale: { 
                format: 'YYYY/MM/DD' 
        }, 
        startDate: moment().subtract(start_date, 'days'), 
        ranges: { 
           'Last 7 Days': [moment().subtract(6, 'days'), moment()], 
           'Last 30 Days': [moment().subtract(29, 'days'), moment()], 
           'Last 45 Days': [moment().subtract(45, 'days'), moment()], 
           'Last Month': [moment().subtract(1, 'month').startOf('month'), moment().subtract(1, 'month').endOf('month')] 
        } 
    }, cb); 

```
It works fine in Google chrome but in Firefox its a bit distorted.

dangrossman:
How's it look when you remove your custom stylesheet?

rksaxena:
I have only bootstrap class "form-control" associated with the element/page. I tried to look the daterangepicker css and found that if I remove ```min-width : 32px```, the dates are rendered properly.

``` css
.daterangepicker .calendar td {
    white-space: nowrap;
    text-align: center;
    /* min-width: 32px; */
}
```

rksaxena:
I

dangrossman:
Here's a screenshot from Firefox taken just now. Unlike your screenshot, there aren't 3 different fonts nor any overflow. You must have another stylesheet of your own on the page...

<img width="356" alt="ss" src="https://cloud.githubusercontent.com/assets/407706/10373903/3c0b20c6-6dbe-11e5-95af-d8b3c290e272.png">

rksaxena:
I removed all style-sheets except the bootstrap the ones mentioned but I am still getting the same problem. My Firefox version is `40.0.3`
```
<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.2.0/css/bootstrap.min.css"/>
<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css"/>
[...]

PR #826, opened 2015-10-08
remove min-width to fix ff border-box issue

peter-mouland:
fixes https://github.com/dangrossman/bootstrap-daterangepicker/issues/825

tested in chrome + FF + ie11
ie10 : width changes to 28px

What browsers are supported and do older browsers need to be 'pixel-perfect'?

### DIFF
--- a/daterangepicker.css
+++ b/daterangepicker.css
@@ -136,11 +136,9 @@
 }
 
 /* Calendars */
-
 .daterangepicker .calendar th, .daterangepicker .calendar td {
   white-space: nowrap;
   text-align: center;
-  min-width: 32px;
 }
 
 .daterangepicker .calendar-table {


dangrossman:
What is the "ff border-box issue"? The date cells are no longer square in Chrome with this change, which makes it look bad. It's supposed to at least vaguely remind people of a wall calendar, where each day lives in a square.

peter-mouland:
I read about border-box not being applied to firefox properly.  This sounded like the case when you see there the width of the cells are 10px too wide in firefox.

If this has an undesired affect on other browsers it makes sense to find another sollution

dangrossman:
#825 doesn't say anything about border-box and these table cells don't have a border.

The solution for a custom stylesheet messing with padding of the calendars is to fix the stylesheet, typically.

PR #838, opened 2015-10-16
Ensure that when window size changes, dropdown stays with input element

mperdeck:
I am using bootstrap-daterangepicker on a responsive site, that will be used by mobile and tablet users as well as desktop users.

I found that after you've dropped down the calendars, when you then resize the window, the dropdown detaches from the input element. Same on a mobile device when you change its orientation.

This is readily reproducable on your page
http://www.daterangepicker.com/

This pull request fixed that issue. When the dropdown is shown (show function is called), a tiny handler is bound to the window.resize event. This calls move when the window resizes, which moves the dropdown back to the input element. When the dropdown is closed (hide function is called), the handler is removed.

By the way, thanks for spending the time and energy on this project. Much appreciated.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -54,6 +54,7 @@
         this.timePickerSeconds = false;
         this.linkedCalendars = true;
         this.autoUpdateInput = true;
+        this.useHtml5Calendar = false;
         this.ranges = {};
 
         this.opens = 'right';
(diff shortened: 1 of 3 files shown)


mperdeck:
I have sent you a second pull request. This introduces a new boolean option useHtml5Calendar. Github seems to treat my two pull requests as one though, so describing it here.

When a mobile user opens the dropdown with the 2 calendars, those calendars take so much space that they often cannot be displayed fully. As a result, picking the dates requires the user to scroll up and down. And it looks a bit weird I think.

To solve this, I introduced the boolean option useHtml5Calendar (default: false). When true, 
* the calendars are not shown in the dropdown;
* the two input boxes for the start and end dates are of type "date" instead of "text".

On Android and iOS, tapping a date input field gets the mobile browser to present a user friendly calendar control. This in effect replaces the 2 calendars that work really well on the desktop and tablets, but not so well on mobiles.

On a responsive site, a bit of JavaScript can check the width of the window and set this option if the window is really narrow.

In addition to the JavaScript code, as part of the pull request I also updated the documentation (the web site file and demo.html).

dangrossman:
I don't think that's a good idea. An HTML5 calendar won't honor any of the dozens of different constraints people can place on the calendar with the minDate/maxDate/dateLimit/isInvalidDate/etc options.

mperdeck:
The objective of using the HTML5 date inputs was to make the mobile experience better. I can see the drawbacks that you mentioned though.

I guess that a way around the date input issues would be to allow the user to pick a "wrong" date (which cannot be avoided), but then validate the input and give an error message if it is invalid.

In my view, an alternative to the date input solution would be to show 1 calendar instead of 2 in the dropdown. The one calendar would be used to enter both start date and end date. I think this solution would work well for the user because it takes far less space. But I couldn't work out how to implement this, so I came up with the HTML5 date input solution.

dangrossman:
The best way is probably to switch to two inputs, for the start and end dates of your range, with two single date pickers. It's already more common to see that than one input with linked calendars (e.g. all the big hotel booking sites have separate inputs for checkin and checkout dates).

dangrossman:
I've added only the code to reposition the picker when the window is resized. Thanks for the suggestion.

    $(window).on('resize.daterangepicker', $.proxy(function(e) { this.move(e); }, this));

PR #845, opened 2015-10-20
additional styling for those wanting to use it without bootstrap

murb:
Not sure whether this is within scope of this project, but was able to use the daterangepicker with just a few additional css tweaks and no bootstrap (still using font-awesome, which could also easily be replaced).


jamesdixon:
@murb this is great. just what I was looking for.

PR #849, opened 2015-10-23
Allow for user to specify an array of disabled days by day index. All…

JamieDixon:
Allow for user to specify an array of disabled days by day index. Allows for user to disable weekends [0, 6] for example.


dangrossman:
You can already do this with the `isInvalidDate` option.

````javascript
$('input').daterangepicker({
  isInvalidDate: function(date) {
    return (date.day() == 0 || date.day() == 6) ? true : false;
  }
});
````

PR #850, opened 2015-10-26
Added dateCellClasses to customize classes for date cells

jackyjieliu:
I added a new option 'dateCellClasses' to customize the classes for individual date cells.

If an array is passed in, it will add the classes in the array to every date cells.
If a function is passed in, the function will be passed the current date for the date cell and the returned array of classes will be set to the date cell.

The main use case is the function. It will allow us to style individual cells.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -255,6 +255,10 @@
         if (typeof options.isInvalidDate === 'function')
             this.isInvalidDate = options.isInvalidDate;
 
+        if (typeof options.dateCellClasses === 'function' ||
+            $.isArray(options.dateCellClasses))
+            this.dateCellClasses = options.dateCellClasses;
+
         // update day names order to firstDay
         if (this.locale.firstDay != 0) {
             var iterator = this.locale.firstDay;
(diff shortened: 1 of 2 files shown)


scotthovestadt:
+1

nnattawat:
+1

jackyjieliu:
@dangrossman Seems like there are some interests in this pull request. Can you accept it if you think the changes are fine?

daterangepicker.js
```diff
@@ -811,6 +815,16 @@
...
                         classes.push('in-range');
 
+                    // Set custom classes configured
+                    var cellClasses;
+                    if (typeof this.dateCellClasses === 'function')
+                        cellClasses = this.dateCellClasses(calendar[row][col]);
+                    else 
+                        cellClasses = this.dateCellClasses;
+
+                    if ($.isArray(cellClasses))
```
nnattawat: Rather than using array, I think it's better to use normal String classes as in [jQuery.addClass()](https://api.jquery.com/addclass/). such as `"class1 class2"`.

jackyjieliu: I used array because classes are arrays in the caller function. Don't really have a preference personally. I guess its up to the owner of the repo

dangrossman:
This has been merged as a new option `isCustomDate` which is a function that takes a date, and returns a string or array of class names.

Issue #855, opened 2015-10-29
The picker doesn't allow to change the start date or end date only


M1chae1:
The picker only allow user to pick the start date and then end date by sequence.

![daterangepicker](https://cloud.githubusercontent.com/assets/5714586/10813287/1528de70-7e59-11e5-9108-bfcc09d6286b.gif)

Does any way allow the user to change the start date or end date only?

set "linkedCalendars": false doesn't solve the problem

dangrossman:
There is not a way to allow the user to change only the start or end date.

PR #858, opened 2015-10-29
Allow single date select

allanwiese:
Property allowSingleSelect created in order to, if enabled, allow selecting just one date (when neither singleDatePicker nor autoApply are enabled).
In clickDate function, if the above condition is true, setStartDate and setEndDate will be invoked based on the calendar being clicked (through cal.hasClass()).
I also created the function updateTime(date, side) that is called when timePicker is enabled. The body of this function is the code that previously existed twice in clickDate().

The reason why I created this pull request is because in previous versions (1.3.5) it was possible to update only one date on the calendars. And there was an issue (#855) about the same matter.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -54,6 +54,8 @@
         this.timePickerSeconds = false;
         this.linkedCalendars = true;
         this.autoUpdateInput = true;
+        // Property to define wheter is possible to alter only one date (if is neither singleDatePicker nor autoApply)
+        this.allowSingleSelect = false;
         this.ranges = {};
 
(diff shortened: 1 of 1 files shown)


dangrossman:
Thank you for sharing, but 

(1) I don't wish to revert this behavior even as a setting

(2) It's broken

E.g. consider when both the start and end of the selected range are in the same month. This patch won't let you choose both the start and end dates in the same month when linkedCalendars is enabled (the default) and will allow you to choose nonsensical ranges like October 30th to October 29th.

ss

PR #863, opened 2015-11-02
Current date and today fixes

ayanko:
Current date and today should be calculated in calendar timezone nor in browser system timezone.

* You should avoid using `moment` without timezone
* You should avoid using `new Date`

They both calculates time using browser offset. And for proper :copyright: calculation we need timezone data.


ayanko:
Reason?

dangrossman:
This library does not need to, and never will, touch time zones. It has caused unending problems in the past. You can set time zones globally on the moment object.

PR #864, opened 2015-11-03
If endDate is clicked => time = 23:59:59

antonellopasella:
When you select a date with timepicker, the time should be the end of the day, not the beginning

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -1282,6 +1282,11 @@
                     }
                     var minute = parseInt(this.container.find('.right .minuteselect').val(), 10);
                     var second = this.timePickerSeconds ? parseInt(this.container.find('.right .secondselect').val(), 10) : 0;
+                    if(hour == 0 && minute == 0 && second == 0){
+                      hour = 23;
+                      minute = 59;
+                      second = 59;
+                    }
                     date = date.clone().hour(hour).minute(minute).second(second);
                 }
                 this.setEndDate(date.clone());


dangrossman:
The time should be whatever the hour/minute/second dropdowns indicate.

antonellopasella:
When you choose a startDate, the endDate is disabled and reset to 00:00:00.
When you click the endDate (day), the dropdowns are enabled, but still 00:00:00

So, if you choose 12/06 and then 14/06 the result is
2015-06-12 00:00:00 to 2015-06-14 00:00:00

IMHO this is wrong, isn't?

dangrossman:
Why do you have the time picker enabled if you don't want the user to choose the time *they* want?

antonellopasella:
IMHO if I click 20/11/2015 I want to select the whole day.
At the moment I just select 20/11/2015 00:00:00, in fact I selected the end
of 19/11/2015

I'm wrong?

2015-11-03 16:24 GMT+01:00 Dan Grossman <notifications@github.com>:

> Why do you have the time picker enabled if you don't want the user to
> choose the time *they* want?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dangrossman/bootstrap-daterangepicker/pull/864#issuecomment-153387349>
> .
>



-- 
Antonello Pasella http://www.pasella.it

sitelifters:
I agree with antonellopasella on this one. If you want to give users the option to select hours, but not require it, then when they select the end date without selecting an hour, it should default to the end of the day.

dangrossman:
If it's a scheduling app and you're choosing the time for a doctor's appointment, does changing their selection to 23:59:59 make sense there? What about the dropoff time for a car reservation? Your preference for a "default", of which there is currently no such thing, is just that, a preference.

antonellopasella:
Agreed but, 00:00:00 is a default preference too 😀

Il giovedì 3 dicembre 2015, Dan Grossman <notifications@github.com> ha
scritto:

> If it's a scheduling app and you're choosing the time for a doctor's
> appointment, does changing their selection to 23:59:59 make sense there?
> What about the dropoff time for a car reservation? Your preference for a
> "default", of which there is currently no such thing, is just that, a
> preference.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dangrossman/bootstrap-daterangepicker/pull/864#issuecomment-161793997>
> .
>


-- 
Antonello Pasella http://www.pasella.it

dangrossman:
There's no code that sets the time to 00:00:00 when the time picker is enabled. There is no default.

PR #900, opened 2015-12-16
Add option to show ISO week of the year

ljank:
We're using ISO-based week numbers in our system which conflicts with locale-based week numbers used in datepicker. Here's what one can get having this handy option:

image

Thank you for such a nice lib! The only missing piece are tests :(


stephanlindauer:
:+1:

dangrossman:
This has been added via the option `showISOWeekNumbers`

PR #904, opened 2015-12-21
Make template options to be complied jquery object

nnattawat:
I have used `daterangepicker` with the custom template that contain `angular` directive. To make angular binding in my template work, it has to be compiled (to be jQuery object) using `$compile` from angular as `{ template: $compile(templateSring)(scope) }`.

I created this PR as it might useful for those who using this `daterangepicker` with Angular app.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -96,7 +96,7 @@
         options = $.extend(this.element.data(), options);
 
         //html template for the picker UI
-        if (typeof options.template !== 'string')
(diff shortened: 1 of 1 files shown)


dangrossman:
Reverted. This patch broke everything when not template is provided (which is the normal/default).

nnattawat:
Ohh I just realised. It has to be `!(options.template instanceof jQuery)`.

PR #915, opened 2016-01-07
Always show calendars when ranges are used

pmacmillan:
The default behaviour with ranges is to only show the predefined ranges and only show the calendar when custom range was selected.

We've started using this feature but want to always show the calendars, even for custom ranges. This option allows us to do that. Example in demo.html.

Default behaviour is maintained.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -365,7 +365,11 @@
             }
         }
 
-        if (typeof options.ranges === 'undefined' && !this.singleDatePicker) {
+        if (options.alwaysShowCalendars) {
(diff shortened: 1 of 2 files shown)


dangrossman:
Thanks for the PR. I generally only add option when multiple people have asked for them, since every option adds more complexity and more to maintain forever. Other people haven't asked for this, so I'm passing on this today.

maruiz93:
I'd like to have this feature added as well.
I think it looks less confusing if you always see the calendars with the selected ranges.

PR #929, opened 2016-01-15
Add additional option for single date selection

aaronmarruk:
Firstly, many thanks for this brilliant plugin.

The team I'm working with have a requirement for selecting a single date when 2 pickers are visible – i.e. when singleDatePicker: false.

I'm proposing adding another option, e.g:
    
    singleDate: true 

which allows for the selection of a single date from multiple pickers. 

What this does is simply, upon date selection, sets the end date equal to the start date and hides the picker.

We would find this a very useful addition to this great plugin (this functionality is a requirement on a number of sites we're building). If there is anything else I need to do to get this implemented, please let me know.

Thanks again.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -46,6 +46,7 @@
         this.dateLimit = false;
         this.autoApply = false;
         this.singleDatePicker = false;
+        this.singleDate = false;
         this.showDropdowns = false;
         this.showWeekNumbers = false;
         this.timePicker = false;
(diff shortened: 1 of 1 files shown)


dangrossman:
Thanks for the PR. Because this is not something I've ever seen on the web, and not something anyone else has asked for in 4 years, I will not be merging and maintaining this option.

aaronmarruk:
Ok, thanks for taking the time to review.

PR #930, opened 2016-01-15
Trigger clickDate event on element

aaronmarruk:
Hello again, 

I'd like to suggest adding a 'clickDate.daterangepicker' event so that it is possible to listen for this on the daterangepicker object.

I believe this would be a useful event to listen to. This is all that's really needed to enable me to achieve the desired outcome, as requested here #929. I'm sure there would be many other uses for this event, and would likely help others.

Thanks for taking the time to review.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -1327,7 +1327,8 @@
             }
 
             this.updateView();
-
+            this.element.trigger('clickDate.daterangepicker', this);
+            
         },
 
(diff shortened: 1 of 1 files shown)


dangrossman:
This event would be fired in the middle of making a selection, a time when the state of the picker should not be used.

aaronmarruk:
Ok, thanks again for reviewing.

PR #932, opened 2016-01-19
Add autoApplyRange and outsideClickCancels options

BlaMaeda:
Hi, I added this two options to the plugin and I thought they may be useful for others.

The first option is to configure the behavior when clicking a range (auto-apply when clicking a range or not). The second option is to make the outside click behave like a cancel.

The defaults for both options are set to behave in a backward-compatible way.

Cheers.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -55,6 +55,7 @@
         this.linkedCalendars = true;
         this.autoUpdateInput = true;
         this.autoApplyRange = true;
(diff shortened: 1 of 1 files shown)


dangrossman:
Thanks for sharing. I'll point people to this PR if someone else ever asks about doing the same. Hasn't happened in the past four years IIRC, so I'd rather not add more options only used by one person.

zhangdaren:
thx

PR #942, opened 2016-01-22
Add functions setMinDate and setMaxDate

MartialSeron:
I don't want to use a "range picker", I need 2 fields : start and end.
I need those 2 functions to update dynamically minDate and maxDate.

Usage : 
```javascript
$('#start').on('apply.daterangepicker', function(ev, picker) {
  $('#end').data('daterangepicker').setMinDate(picker.startDate);
});

$('#end').on('apply.daterangepicker', function(ev, picker) {
  $('#start').data('daterangepicker').setMaxDate(picker.startDate);
});
```

Works well for me.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -441,6 +441,42 @@
 
         constructor: DateRangePicker,
 
+        setMinDate: function(minDate) {
+            if (typeof minDate === 'string')
+                this.minDate = moment(minDate, this.locale.format);
+
+            if (typeof minDate === 'object')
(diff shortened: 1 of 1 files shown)


dangrossman:
Thank you for the PR, but I don't think this is a common enough use case to necessitate additional code to maintain. It can also be achieved with no code changes by recreating the picker with the new `minDate` or `maxDate` value, the same way it was initially created. You can call `.daterangepicker()` on an element any number of times, and the old picker will be replaced by a new one.

PR #946, opened 2016-01-27
Add method `updateMomentLocale` to translate calendar view

nnattawat:
I have added method called `updateMomentLocale` to translate calendar view to the given moment locale key

```javascript
// The usage from the demo would be
$('#config-demo').data('daterangepicker').updateMomentLocale('fr')
```

I have tested and the demo page and it worked fine for me. If you thing it would be good to provide this feature, I will add an example on the `demo.html` page.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -1507,6 +1507,20 @@
             this.container.remove();
             this.element.off('.daterangepicker');
             this.element.removeData();
+        },
+
+        updateMomentLocale: function(key) {
+          moment.locale(key);
+          this.locale.daysOfWeek = moment.weekdaysMin();
+          this.locale.monthNames = moment.monthsShort();
+          this.locale.firstDay = moment.localeData().firstDayOfWeek;
+          for (var attr in this) {
+            if (this[attr] && this[attr] instanceof moment().constructor) {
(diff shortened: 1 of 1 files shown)


dangrossman:
The picker already pulls that data from moment, so why can't you set the locale on the global moment object before creating it?

nnattawat:
My use case is to change the language after calendar is initialised. So I add this method instead of change `moment.locale()` and then re-initialise calendar again.

dangrossman:
Good point, you can also change the locale and re-initialize the calendar. That seems like a much better option than maintaining code that'll likely never be used by anyone else. :+1:

shihao316558512:
@nnattawat I used your code,it works verry well.``

juliomac:
It would be great to have this Pull merged. I too have the same use case of @nnattawat . I believe MANY other also do.

dangrossman:
I disagree that this would be great to have, or that many people need it. I believe zero people need it, since you can update your picker to reflect a new locale without any new function.

Just do this:

````
moment.locale('whatever');
$(yourelement).daterangepicker(options);
````

Any time you call `.daterangepicker()`, the picker is recreated from scratch, which would pick up any changes to your options or to the global moment object.

PR #952, opened 2016-02-02
Use strict validation on formInputsChanged

raulpopadineti:
Don't use the permissive `moment(date, format).isValid()` function, use `moment(date, format, true).isValid()` instead. 

This won't allow manually adding inputs like `1/1/2016` when using the expanded form like `Fri, Jan 1 2016`. This way the plugin will become less error prone.

When an input is invalid we revert back to the previous input value.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -1446,9 +1446,9 @@
                 this.setEndDate(end);
 
                 if (isRight) {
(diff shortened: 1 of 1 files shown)


raulpopadineti:
@dangrossman Sorry, I missed one case. I'll open another PR for this.

PR #953, opened 2016-02-02
Use strict validation on formInputsChanged

raulpopadineti:
Don't use the permissive `moment(date, format).isValid()` function, use `moment(date, format, true).isValid()` instead.

This won't allow manually adding inputs like `1/1/2016` when using the expanded form like `Fri, Jan 1 2016`. This way the plugin will become less error prone.

When the end input is before the start input, we change the start input as well.
When the start input is after the end input, we change the end input as well.
When an input is invalid, we revert back to the previous input value.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -1434,8 +1434,8 @@
 
         formInputsChanged: function(e) {
             var isRight = $(e.target).closest('.calendar').hasClass('right');
-            var start = moment(this.container.find('input[name="daterangepicker_start"]').val(), this.locale.format);
-            var end = moment(this.container.find('input[name="daterangepicker_end"]').val(), this.locale.format);
+            var start = moment(this.container.find('input[name="daterangepicker_start"]').val(), this.locale.format, true);
+            var end = moment(this.container.find('input[name="daterangepicker_end"]').val(), this.locale.format, true);
 
             if (start.isValid() && end.isValid()) {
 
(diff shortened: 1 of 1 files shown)


raulpopadineti:
@dangrossman let me know when you got time to review it. Thanks!

dangrossman:
I don't know that adding `true` to the moment constructor won't break websites relying on the current behavior. You also duplicated a lot of logic that already exists in the `setStartDate`/`setEndDate` functions.

raulpopadineti:
The logic is not duplicated. First if you want to write in an input `1/1/2016` as the plugin is now, it will return to you a date from year `0015` if you don't have a `minDate` set because this:

```

**Comment by raulpopadineti** [comment]
The logic is not duplicated. First if you want to write in an input 1/1/2016 as the plugin is now, it will return to you a date from year 0015 if you don't have a `minDate` set because this:

```
formInputsChanged: function(e) {
     var isRight = $(e.target).closest('.calendar').hasClass('right');
     var start = moment(this.container.find('input[name="daterangepicker_start"]').val(), this.locale.format);
     var end = moment(this.container.find('input[name="daterangepicker_end"]').val(), this.locale.format);
     ...
```

is too permissive. You don't actually validate the date. It's written in the momentjs docs:

```
Moment's parser is very forgiving, and this can lead to undesired behavior.
...
As of version 2.3.0, you may specify a boolean for the last argument to make Moment use strict parsing. Strict parsing requires that the format and input match exactly, including delimeters.
```

Second you don't even treat the invalid inputs in the `formInputsChanged()` function. Of course the `isValid()` will never return invalid the way it's written now.

I'm afraid you haven't even take a good look over the code, but thanks for the response.

PR #956, opened 2016-02-04
Added boolean "showWithRanges" option for show always picker with ranges

arturmamedov:
When ranges{} object is defined picker not shown, because show-calendar class is removed, and when selecting a range picker are hide. This is not usual for people and for avoid this i add "showWithRanges" that are default false, but if setted to true avoid the hideCalendars() on clickRange()

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -335,6 +335,9 @@
             }
             if(this.showWithRanges === false)
               list += '<li>' + this.locale.customRangeLabel + '</li>';
+            else
+              list += '<li style="display: none;">' + this.locale.customRangeLabel + '</li>';
+            
             list += '</ul>';
             this.container.find('.ranges').prepend(list);
         }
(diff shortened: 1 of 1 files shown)


pane-e-design:
+1

maoueh:
Yep, I thought I had a bug at first because the calendar was not always displayed. Seemed like this was a feature after all :)

Would love to have it configurable because in my case, I prefer to always have it on.

arturmamedov:
And if on rangeClick() we can remove the auto apply, and see the selecting date apply in picker, it would be the top! How in ga in example.

webdevsega:
+1 It is necessary

dangrossman:
This has been added as an option with the name `alwaysShowCalendars`. Thanks for the suggestion.

arturmamedov:
Yes i try `alwaysShowCalendars` but its now work, or its not work how i expected.

dangrossman:
It's exactly the same as the code you wrote, just with a different name. You'll need to explain yourself if you want help.

arturmamedov:
I wont that the date picker is always show! 
I now try in [jsFiddle ](https://jsfiddle.net/arturmamedov/wmo79sdn/2/) and its work ....  when i develop on local machine the show always not work and show me only the ranges... Thanks! All work perfectly!

PR #959, opened 2016-02-05
Fire hide event after view gets updated

nnattawat:
I have an method bound with `hide.daterangepicker` event but I have to read the content in the view on my methods. I found out that `hide.daterangepicker` is not getting the latest view. This PR is to move `apply` and `hide` events to fire after the view get updated. I thinks it make mode sense this way.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -1281,6 +1281,8 @@
             // * if single date picker mode, and time picker isn't enabled, apply the selection immediately
             //
 
+            var needApplying = false;
+
             if (this.endDate || date.isBefore(this.startDate, 'day')) {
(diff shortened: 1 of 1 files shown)


dangrossman:
Why are you reading "the content in the view"?

nnattawat:
I did not really read the content. What I did is to move start and end date inputs out of the calendar and I need to remove `.active` class from the inputs once the calendar is close but I cannot do it on `hide` event as the view will get overridden after.

screen shot 2016-02-05 at 3 17 41 pm


It's a bit custom and rare. It should not make any different in most of the cases but I still think it is more correct.

PR #960, opened 2016-02-05
Fire hide event after cancel and apply

nnattawat:
This is related to https://github.com/dangrossman/bootstrap-daterangepicker/pull/959.

It visually makes more sense to trigger `apply` and `cancel` events before `hide` as you really click apply/cancel before the calendar is hidden.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -1331,15 +1331,15 @@
         },
 
         clickApply: function(e) {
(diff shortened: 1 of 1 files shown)


dangrossman:
All the `hide` method does is clean up some internal state and fire a callback if provided. The order in which these statements are run is inconsequential.

PR #961, opened 2016-02-05
disableHoverDate

s2oBCN:
The hoverDate function I found it confusing that the input field values are changing as I hover the calendar. Specially when changing hours.
So  I added a new option: "disableHoverDate", that indicates whether hover the calendar should automatically change the displayed value of an <code>&lt;input&gt;</code> element.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -54,6 +54,7 @@
         this.timePickerSeconds = false;
         this.linkedCalendars = true;
         this.autoUpdateInput = true;
+        this.disableHoverDate = false;
         this.ranges = {};
 
         this.opens = 'right';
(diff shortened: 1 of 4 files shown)


dangrossman:
Thank you, but I am passing on this PR.

mkoch42:
Would be great if you could commit this one.  Our goal is to clear either the start or end field and just require one date for some pages where we want two date fields, but want to use all of the great functionality of your tool including the 'ranges', but want to allow user to clear one field or the other should they choose to not want to limit it to a range.  

I got this working with simple event listening on a clear glyphicon added to the dom after drp loads but then when you mouseover on your way to click Apply it fills in the date.  If we could disableHoverDates so that the fields do not update, it is easy to clear and then click 'Apply'. (autoUpdateInput will be false) so it would be great to add this code to the master for this purpose at least.

Thanks for re-considering!
Matt

PR #967, opened 2016-02-11
Ranges with subranges

ShadesOB:
First up, very sorry I had an old copy of dangrossman's code. Also sorry I'm too lazy to update mine to latest copy. So feel free to use it or simply use the idea.  Or just ignore this. I won't be upset. :p

I needed to have more options on the ranges. So I changed it to allow below example.
This requires bootstrap css and jquery. But I'm sure you can change those if wanted.

(code below is php. But you should get the idea)
<pre>
'ranges' => [
                                'Q1' => [
                                    "2016" => [
                                        date('d/m/Y', strtotime('2016-01-01')),
                                        date('d/m/Y', strtotime('2016-04-01 -1day'))
                                    ],
                                    "2015" => [
                                        date('d/m/Y', strtotime('2015-01-01')),
                                        date('d/m/Y', strtotime('2015-04-01 -1day'))
                                    ],
....
                                ],
                                'Q2' => [
                                    "2016" => [
                                        date('d/m/Y', strtotime('2016-04-01')),
                                        date('d/m/Y', strtotime('2016-07-01 -1day'))
                                    ],
                                    "2015" => [ 
....
<pre>

image

### DIFF
--- a/daterangepicker.css
+++ b/daterangepicker.css
@@ -9,367 +9,366 @@
 /* Container Appearance */
 
 .daterangepicker {
-  position: absolute;
-  background: #fff;
-  top: 100px;
-  left: 20px;
-  padding: 4px;
-  margin-top: 1px;
-  border-radius: 4px;
-  width:278px;
+    position: absolute;
+    background: #fff;
+    top: 100px;
+    left: 20px;
+    padding: 4px;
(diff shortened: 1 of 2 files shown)


dangrossman:
Thanks for sharing.

Issue #984, opened 2016-02-22
When maxDate is set to current month, opening a custom range should show previous and current month VS current and next (which is invalid)


TheDartCode:
When we have linkedCalendars and maxDate set to the current month, and select "custom range", the two calendars that open up show the current and next month.
However, since maxDate is in the current month, it would be more useful to show  the previous and the current month.

Proof: https://jsbin.com/holagob/1/edit?html,css,js,output

tim-u:
+1, I noticed it too.

PR #985, opened 2016-02-22
Fix issue in linkedCalendars, where if the maxDate is this month, the…

TheDartCode:
… right calendar showed a month after the maxDate prop.

Solves #984

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -547,6 +547,11 @@
                     this.rightCalendar.month = this.startDate.clone().date(2).add(1, 'month');
                 }
             }
+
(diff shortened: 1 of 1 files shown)


dangrossman:
This was reverted because it introduced new breaking bugs.

Reproduction steps, with your code:
1. Set maxDate to 2/29/2016
2. Click the left arrow to scroll back until December 2015 is in view
3. Click on a date in December 2015

The calendars unexpectedly update to display January and February 2016, so the selection in December cannot be completed.

TheDartCode:
You were right, I was erroneously comparing rightCalendar.month to maxDate.month() when I should be comparing it to maxDate itself. Fixed in #989

Thanks for your instant response!

PR #989, opened 2016-02-23
Fix issue in linkedCalendars, where if the maxDate is this month, the...

TheDartCode:
… right calendar showed a month after the maxDate prop.

Solves #984

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -324,7 +324,7 @@
                 // after the maximum, don't display this range option at all.
                 if ((this.minDate && end.isBefore(this.minDate)) || (maxDate && start.isAfter(maxDate)))
                     continue;
-                
+
                 //Support unicode chars in the range names.
                 var elem = document.createElement('textarea');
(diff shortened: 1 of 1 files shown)


maxailloud:
I would be interested by this fix.

I'm using through https://github.com/fragaria/angular-daterangepicker and with your fix I have an error.
Don't know why but I have false in maxDate sometimes, so error when using clone method.
Adding 'this.maxDate && ' to the condition make it work.

TheDartCode:
@maxailloud  The wrapper library you mention sets maxDate to false if none is provided (check line coffee/angular-daterangepicker.coffee#L146).
Don't know if it is the right approach to set false to that field, but in order to cover the case I added the check, too.

Cheers!
Hope the PR is accepted :)

floydpraveen:
was looking for this, +1  for this PR

maxailloud:
Any news here?

dangrossman:
This code created new issue #1098

PR #990, opened 2016-02-23
Added Function updateInputText

ramilani12:
Some Angular Plugins still use function updateInputText, older versions of bootstrap-daterangepicker

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -1518,6 +1518,10 @@
                 this.element.trigger('change');
             }
         },
+        
+        updateInputText: function(){
+          this.updateElement();  
+        },
 
         remove: function() {
             this.container.remove();


dangrossman:
I'm not sure what this is about, but this is not part of the public API for this date picker. It's an internal helper function and doesn't need to be exposed.

ramilani12:
Hi Dan
I think, I was not sure.
Some implementations of angular that using bootstrap-daterangepicker were using older versions of plugin.
For example this plugin(https://github.com/luisfarzati/ng-bs-daterangepicker) was using the version : 1.3.21 of daterangepicker. Then the implementation was calling function  updateInputText if you try to update to the newest version of plugin it won't work because the function does not exist anymore.

dangrossman:
Either continue using 1.3.21, or update your plugin to use the new version correctly.

ramilani12:
Yes, Dan . Of course.
I've just updated the plugin.
By the away the implementation is not mine. :)

PR #994, opened 2016-02-28
Add new Events updateView

zocek:
Adding new events that will allow a change rendering the calendar.


dangrossman:
What is the purpose of this? Triggering an event in response to internal helper functions firing invites people to rely, and build on, internals that are subject to change. This isn't part of the public API.

zocek:
Allowing customization rendering the calendar. For example, custom color selected dates in the calendar, if they can not use the (occupied terms). If you see a better solution, so please proposal.

PR #1001, opened 2016-03-03
Callback Not Called When Clicking Out of Input

pixelshaded:
This ensures that formInputsChanged will be called before hide in a case where you click outside the control right after changing an input value (but without "submitting it")


dangrossman:
This causes bugs when there are multiple date pickers on a page. It allows multiple pickers to be opened then they never close when clicking elsewhere on the page.

pixelshaded:
So debugging, this problem is caused by the deregistering of events in hide. To reproduce, open a calendar, then open another calendar on the page.

Previous Flow
Outside Click Event
Hide and Remove Events
Show New Calendar (Add Events)

Timeout Flow
Outside Click Event
Show New Calendar (Add Events)
Hide and Remove Events

So in the timeout flow, events in the daterangepicker namespace are removed after the show. Since the control this.isShowing is true, new calls to show won't add events. Since we've lost events, outside click won't cause the shown control to close.

pixelshaded:
Put both show and hide logic within a timeout. This ensures they execute in the proper order relative to each other (i.e. to handle use case when one is being hidden and another shown from a single click), but ensures they get put on the stack after formInputChanged. This fixed both the callback issue and multiple controls remaining open on the page. @dangrossman you have more experience with the source code to verify whether or not these timeouts might have some unclear drawbacks / consequences. Briefly running through all the controls on the test website locally didn't reveal anything obvious.

Created a variable for 'this' instead of using bind since these functions get called often and closures are faster then binds.

JamesHenry:
Any update on this either way @dangrossman?

Issue #1008, opened 2016-03-08
When turn on timePicker get message about 'Deprecation' from moment.js


KirillMehtiev:
What it can be? 

![image](https://cloud.githubusercontent.com/assets/10028598/13604017/7016c426-e549-11e5-8820-e3bf34abccbd.png)

dangrossman:
There's nowhere I parse a string into a date without a format, so I don't know what the cause of the warning is.

KirillMehtiev:
Thanks for replay!

Issue #999, opened 2016-03-03
moment.js deprecation warning


jbruni:
This is appearing at console when I use bootstrap-daterangepicker:

> Deprecation warning: moment().subtract(period, number) is deprecated. Please use moment().subtract(number, period).
> Deprecation warning: moment().add(period, number) is deprecated. Please use moment().add(number, period).

dangrossman:
Fix merged

rafaeldruiz:
I'm still receiving this warning with the last version of daterangepicker.

 * Minified by jsDelivr using Terser v3.14.1.
 * Original file: /npm/daterangepicker@3.0.5/daterangepicker.js

Deprecation warning: moment().subtract(period, number) is deprecated. Please use moment().subtract(number, period). See http://momentjs.com/guides/#/warnings/add-inverted-param/ for more info.

Any news?

dangrossman:
It was fixed 3 years ago, which is the age of this issue you're commenting on.

Feel free to search the file: https://github.com/dangrossman/daterangepicker/blob/master/daterangepicker.js

There are no instances of moment().subtract(period, number) in the code.

rafaeldruiz:
Sorry for my mistake, the error was coming from another plugin.

Issue #993, opened 2016-02-25
Moment.js Deprecation Warning


zthomas1:
When the picker is first expanded and the user hover's over "Custom Range" line 871 creates the following warning:

Deprecation warning: moment construction falls back to js Date. This is discouraged and will be removed in upcoming major release. Please refer to https://github.com/moment/moment/issues/1407 for more info.

dangrossman:
Are you using an old copy of this library?

zthomas1:
I was using an older version but, i updated to the latest 2.11.2 and the warning persists.
Thank you for responding!

dangrossman:
Current version is 2.1.18 ...

zthomas1:
I meant that I update my moment.js to 2.11.2, I was using picker 2.1.17, 2.1.18 does not seem to show a warning.  Just when I thought I had a real issue, you already beat me to it.

Hats off Sir.

dangrossman:
2.1.17 didn't have this issue either. It hasn't existed in, oh, a year or so?

PR #1010, opened 2016-03-09
Don't do date comparison if we don't have a maxDate

djpc123:
This should fix moment.js warnings some people are seeing in #1008,  #999 and #993


dangrossman:
Thanks, I merged this manually

PR #1018, opened 2016-03-16
Autoapply fix

mrdevguy:
When autoApply is set to false, I would expect the behaviour to be consistent for all range selections. However it immediately hides the popup when you click on predefined date ranges. Now it will only close when you click the Apply button.


dangrossman:
I will not accept a change in default behavior like this. You can't spring that kind of thing on tens of thousands of people, whose websites have worked this way for the past 4 years.

PR #1046, opened 2016-04-14
Fix the module definition

mcfedr:
Export constructor in node/browserify environment


dangrossman:
This puts me in a tough spot. I have no idea if this works correctly, and if it breaks AMD/RequireJS or whatever the people that wrote it that way use. I have none of these environments set up to test.

mcfedr:
I complete understand where you are coming from, and I can see there is some risk associated.

I have tried to not actually change much, but I have exported the `DateRangePicker` constructor in both node and amd where it was not exported before.

I have also tried to simplify it, but kept the 'unusual' bit around jQuery that must have been put there for a reason. Otherwise I tried to stick closely to the [Universal module definition]( https://github.com/umdjs/umd)

mcfedr:
So I can see if works :)

mcfedr:
just so you know, the reason I wanted to do this, I am using this via browserify, and I want access to the constructor so I can adjust it a bit.

mcfedr:
no chance?

PR #1052, opened 2016-04-18
Fix ranges option, hide calendar by default

anilmaurya:
Show calendar on first click only when `alwaysShowCalendars` option is enabled.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -1342,7 +1342,9 @@
           }
           if (customRange) {
               this.chosenLabel = this.container.find('.ranges li:last').addClass('active').html();
-              this.showCalendars();
+              if(this.alwaysShowCalendars){
+                this.showCalendars();
+              }
           }
         },
 


dangrossman:
If "custom range" is already highlighted, then the calendars should be shown. This change makes the UI inconsistent in this situation.

anilmaurya:
@dangrossman On page load `custom range` is highlighted by default therefore calendar is shown even when I don't  want to display calendar.
It start behaving correctly After I click on predefined range once.

dangrossman:
Why don't you want to display the calendars? How is the user supposed to choose the date range they want? Why would they guess that they need to click on the already selected option to make the calendars appear so they can continue? The answer is that they won't guess that, you're creating a confusing interface.

anilmaurya:
@dangrossman I think you didn't understood my question.

Let me try again.

You said `If "custom range" is already highlighted, then the calendars should be shown` and I get it.
Now I don't want `custom range` to be highlighted by default.

How do you propose to achieve it.

dangrossman:
Set a `startDate` and `endDate` (initial selected date range) that match one of your other predefined ranges, then that range will be selected instead.

anilmaurya:
Thank you ♡

PR #1077, opened 2016-05-05
Add functionality to add custom class to dates

brobey8:
Adding a couple of props (**specialDates** - array of moment dates, **specialClass** - class name to be appended) that will add a custom class to any dates specified in the **specialDates** array that is set in the options.

Can easily style them with the custom class that is chosen (falls back to 'special')


dangrossman:
This has been merged as a new option `isCustomDate` which is a function that takes a date, and returns a string or array of class names.

PR #1081, opened 2016-05-10
Fix end time getting reset to start time

### LINKED ISSUE #943: End Time Selection Changing
So it looks like this has been raised before in #895, but I don't think b1bc05d7 fully fixed the issue. See [this](https://jsfiddle.net/8cyw64t1/) jsfiddle with the most recent master. 

As you can see in the gif below, when you select new dates that are within the time frame of the current selection, your time selection stays the same. However, if the new start date is past the current end date, the end time is reverted to 12:00:00.

gif

quainjn:
Fixes #943 by preserving the end time as it was before. Sets the end date to either the new start date or the day after the new start date if the end time is before the start time.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -853,7 +853,9 @@
                     }
 
                     if (selected.isBefore(this.startDate))
-                        selected = this.startDate.clone();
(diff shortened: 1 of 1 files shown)


dangrossman:
After applying this change I was able to make invalid selections (e.g. changed the time on the left to 7AM, the time on the right to 6AM, then click the same date on the calendar twice, resulting in a range of 7AM to 6AM on some date, which is impossible).

PR #1087, opened 2016-05-18
added methods to set minDate and maxDate

mmwillet:
Hey, this is some functionality which we added into our dateRangePicker.

Because we listen for changes to minDate and maxDate, we wanted to be able to update dateRangePicker for those changes without having to reinitialize; so we added a setMinDate method and a setMaxDate method to the dateRangePicker class. 

Please let us know if there is anything that should be changed or if you foresee any issues. Thanks!

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -504,6 +504,62 @@
             this.updateMonthsInView();
         },
 
+        setMinDate: function(minDate) {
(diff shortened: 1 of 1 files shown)


dangrossman:
There is no drawback to reinitializing to change options.

PR #1092, opened 2016-05-23
Extra options to be more flexible

jmsv23:
Option to control de distance of the box and the element.
Option to add minimum selection.
Option to decide if selecting a range apply the selection or just update the calendar.


VeeK727:
Thanks jmsv23 for this. I used your code to add a minimum limit on range. Works perfectly. Although I removed the classes 'off' and 'disabled' as it seems weird or confusing from a non technical person's point of view.

Dan should atleast merge some of it.

+1

PR #1102, opened 2016-06-02
No endDate before startDate

bsa01:
As the handling of left and right calendar was confusing our users, I tried to disable all the day's in the calendar's that are before the startDate, when it comes to the endDate selection.
Hopefully this is fine?

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -790,7 +790,7 @@
                         classes.push('off');
 
                     //don't allow selection of dates before the minimum date
-                    if (this.minDate && calendar[row][col].isBefore(this.minDate, 'day'))
+                    if ((this.minDate && calendar[row][col].isBefore(this.minDate, 'day')) || (!this.endDate && calendar[row][col].isBefore(this.startDate, 'day')))
                         classes.push('off', 'disabled');
 
                     //don't allow selection of dates after the maximum date


dangrossman:
That's fine for you, but I do not wish to change the behavior for everyone else. It's not nice to change how other peoples' websites have worked for 4+ years.

bsa01:
But's that what you already did ("change how other peoples' websites have worked for 4+ years")

dangrossman:
You upgraded to a new major version (1.x to 2.x) voluntarily. If you had a dependency on daterangepicker 1, you wouldn't "npm update" and suddenly have totally different behavior.

bsa01:
No we used bootstrap-daterangepicker-rails gem version 0.1.1 (which had your version 1.3.17 inside)
Version 0.1.2 "suddenly" had your version 2.1.12 and "booom" everything changed ;-)

But it's ok. We will no longer use the gem but a patched version of your lib.
Thanks for your work.

dangrossman:
I didn't make any gem; you can't blame me for that.

kakid:
It was very confusion for our users too, in the version 1.X, they knew that they can change the start date/end without changing the end/start date, that was very frustrating.

Is it possible to add an option, to keep the old selection method ?

dangrossman:
Yes, you can have that option. To enable it, download the old version and use that....

PR #1105, opened 2016-06-06
Functionality to disable minutes in timePicker

aggarwalankush:
timePickerMinutes: (boolean) Show minutes in the timePicker. Default set to true for backward compatibility.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -44,6 +44,7 @@
         this.timePicker = false;
         this.timePicker24Hour = false;
         this.timePickerIncrement = 1;
+        this.timePickerMinutes = true;
         this.timePickerSeconds = false;
         this.linkedCalendars = true;
         this.autoUpdateInput = true;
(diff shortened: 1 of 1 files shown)


dangrossman:
What is the use case for this?

aggarwalankush:
Many projects in my company uses Hadoop and one of the best approaches of storing data in hadoop is in YYYY/MM/DD/HH directory structure. For monitoring dashboards, we use your bootstrap-daterangepicker and we don't have requirements for minute dropdown. That's my use case, I think other companies might have similar use cases. 

`timePickerIncrement: 60` would accomplish the task but user experience would be bad as user will see a dropdown for minutes but can't change it. It would feel like a bug to user.

Right now, we are using following css to accomplish similar task in our projects.
 
```
.calendar-time > div {
    visibility: hidden;
}

.monthselect, .yearselect, .calendar-time > div > .hourselect {
    visibility: visible;
    color: #333;
    background-color: #fff;
    border-color: #ccc;
    width: 50%;
    text-align-last: center;
}

.calendar-time > div:before {
    visibility: visible;
    color: black;
    content: "Hour ";
}

.range_inputs>button {
    width: 80px;
    margin-top: 10px;
}
```
I thought it wouldn't harm to have an option to disable minute dropdown.

Without above CSS - Using `timePickerIncrement: 60`
screenshot 2016-06-07 21 30 54


With above CSS
hourpicker

I feel like second screenshot provides better user experience.

PR #1115, opened 2016-06-09
`autoRange` Feature.  Auto selects specified range

julianiristv:
Added `autoRange` option as a `number`.  If specified, it will ( on `clickDate` event) select the start_date and automatically select an end_date.  The end_date will be `autoRange` days ahead of the start_date.

Ex: User clicks on a date (2015-02-01) while the `autoRange` option is set to 7.  This will automatically select and highlight the daterange 2015-02-08).  Useful for daterangepickers that can only scope to specific date ranges.  i.e. 30 day span (month) or 7 day span (week).

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -132,9 +132,6 @@
 
         if (typeof options.locale === 'object') {
 
-            if (typeof options.locale.autoRange === 'number')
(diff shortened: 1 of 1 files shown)


dangrossman:
Thank you for the suggestion, but I do not wish to add this to the code base at this time.

PR #1120, opened 2016-06-10
Add moment constructor option (idea from vis.js timeline).

daniel-pfeiffer:


daniel-pfeiffer:
Without a timePicker, start and end got nicely rounded to start and end of day.  But with a timePicker this step was skipped, settings seconds to 0 even for the end and leaving µs at whatever they were before changing.  Now rounding goes to days, minutes or seconds depending on the timePicker options.

PR #1124, opened 2016-06-14
fix removal of cutom range button if alwaysShowCalendars

RobertSchouten:


dangrossman:
The list of ranges serves three purposes: allowing the selection of a predefined range, displaying the calendars to choose a custom range, and showing whether the current selection corresponds to one of the ranges in the list or is a custom selection. This last purpose is served by the inclusion of "Custom Range" in the list even when "alwaysShowCalendars" is enabled.

k-funk:
I also find this awkward. If alwaysShowCalendars is true, and I click Custom Range, nothing happens. This "button" isn't actually a button, it's an item on a list that reflects the context of the calendar choices, while the other ranges are buttons that actually perform a function. Dead buttons are bad for UX.

I think it'd be best to hide Custom Ranges until the user actually chooses a custom range on the calendars, or always hide it.

dangrossman:
The option `showCustomRangeLabel` was added 10 months ago

k-funk:
Missed that. Thank you!

PR #1140, opened 2016-06-21
Allowing other date formats and checking if date is valid

multiwebinc:
This allows the user to set start and end dates as a string based either on the Americanized default locale that bootstrap-daterangepicker uses or on any of the formats that moment.js accepts. It also prevents an invalid moment.js object from being set to this.startDate and this.endDate.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -448,8 +448,15 @@
         constructor: DateRangePicker,
 
         setStartDate: function(startDate) {
-            if (typeof startDate === 'string')
-                this.startDate = moment(startDate, this.locale.format);
+            if (typeof startDate === 'string') {
+                var d = moment(startDate, this.locale.format);
+                if (!d.isValid()) {
+                    d = moment(startDate);
+                }
+                if (d.isValid()) {
+                    this.startDate = d;
+                }
+            }
 
             if (typeof startDate === 'object')
                 this.startDate = moment(startDate);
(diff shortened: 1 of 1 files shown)


dangrossman:
If someone passes a string in a non-ISO format without an accompanying format string, it will trigger a deprecation warning, and in future versions of moment.js will simply fail

https://github.com/moment/moment/issues/1407

multiwebinc:
Good point. It says:

> For consistent results parsing anything other than ISO 8601 strings, you should use String + Format.

Is it possible to allow either ISO 8601 strings *or* a given format? If not, I could just remove that part and force the user to either use the default value or provide their own format.

dangrossman:
That would be the state of things already. All adding an extra `isValid()` check here would do is allow website developers to provide bad inputs and, instead of immediately knowing they've done something wrong, silently ignore their input so they're confused if they're even using the API right.

multiwebinc:
As it is now, if an invalid date is provided for `setStartDate` it makes a calendar with a bunch of `NaN` everywhere, however `setEndDate` silently ignores invalid dates so that could lead to confusion as well.

I did not see any way to provide feedback to the developer, so what about something like this using `console.error()`?:

```
if (typeof startDate === 'string') {
    var d = moment(startDate, this.locale.format);
    if (!d.isValid()) {
        console.error("Date string does not match locale.format");
    }
    else {
        this.startDate = d;
    }
}
```

PR #1144, opened 2016-06-23
Fix bug that breaks view when select date range within 24 hours

honmaaax:
Forgive me if my understanding is mistaken.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -322,7 +322,7 @@
 
                 // If the end of the range is before the minimum or the start of the range is
                 // after the maximum, don't display this range option at all.
-                if ((this.minDate && end.isBefore(this.minDate, this.timepicker ? 'minute' : 'day')) 
+                if ((this.minDate && end.isBefore(this.minDate, this.timepicker ? 'minute' : 'day'))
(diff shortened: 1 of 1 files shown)


dangrossman:
What is the bug? What are the steps for me to reproduce it?

honmaaax:
@dangrossman 
I can see the bug when I do as follows.

1. Use `react-bootstrap-daterangepicker`
2. Set `timePicker={true}` and `timePicker24Hour={true}`
3. View daterangepicker component on Chrome
4. Set date range within 24 hours
5. Click a day after endDate on daterangepicker's calender
6. You can see the bug

It fixes by the changes of this pull request.

dangrossman:
I don't know anything about `react-bootstrap-daterangepicker` but I wasn't able to reproduce it using those instructions at http://www.daterangepicker.com/#config

honmaaax:
@dangrossman 
Sorry, I checked that it reproduces the bug at the site as you said.
I should investigate the cause of the bug, so I close this pull request.

honmaaax:
@dangrossman 
https://jsfiddle.net/4nrjLL20/
Please check this code and do as follows.

1. Click the input on preview.
2. Click the date after "2016-06-21"
3. The view of all data is broken

PR #1146, opened 2016-06-27
Fix picker problems

misticini:


misticini:
Thank you for closing this pull request. That was a mistake ;)

PR #1150, opened 2016-06-29
autoUpdateInput currentdate

guerinaxel:
This is a fix for single datepicker : 
if the autoUpdateInput is false, when you click on the current date the field won't update.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -1119,7 +1119,7 @@
             }
 
             //if a new date range was selected, invoke the user callback function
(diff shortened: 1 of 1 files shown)


dangrossman:
Your callback function is only triggered when the selected date is changed. Clicking on the same date that's already selected should not trigger the callback. I do not wish to change this behavior on people in a minor version update posing as a bug fix, sorry. I think that would be wrong.

If you want to write code that reacts to a date being clicked every time, even if the date is the one that's already selected, there's an event for that, `apply.daterangepicker`.  For example, http://www.daterangepicker.com/#ex5

PR #1154, opened 2016-06-30
Add boolean option iniInputWithStartDate

a55660988:
iniInputWithStartDate is boolean option.
Its default value is true, just work like before.
The purpose of iniInputWithStartDate is 
If input is originally populated with date value, 
assign iniInputWithStartDate to false in order NOT to 
change(auto update) the input to StartDate when initialize

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -31,6 +31,7 @@
         //default settings for options
         this.parentEl = 'body';
         this.element = $(element);
+        this.iniInputWithStartDate = true;
         this.startDate = moment().startOf('day');
         this.endDate = moment().endOf('day');
         this.minDate = false;
@@ -164,6 +165,9 @@
 
         }
         this.container.addClass(this.locale.direction);
+
+        if (typeof options.iniInputWithStartDate === 'boolean')
+            this.iniInputWithStartDate = options.iniInputWithStartDate;
 
         if (typeof options.startDate === 'string')
             this.startDate = moment(options.startDate, this.locale.format);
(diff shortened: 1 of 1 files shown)


dangrossman:
You can set `autoUpdateInput` to `false` and update it the way you want in a callback function or event listener.

a55660988:
Yes, that's fine.
It would be a little bit confused that `autoUpdateInput = true` also initially set value.
Just separate `autoUpdateInput`, one is for auto update input, the other is for setting initial value.

dangrossman:
There's no need for this option.

If you want your input to contain the `startDate` when the page loads, then you should set the `startDate` when creating your date range picker, and that's what it will put in your input. 

If you want your input to contain something other than the `startDate` when the page loads, you have two options:

1. Set `autoUpdateInput` to `false`

2. Put whatever you want into your input *after* creating the date range picker

````
$('#yourinput').daterangepicker();
$('#yourinput').val('Initial Value');
````

a55660988:
My input is initialized when rendering page, so I set...
`singleDatePicker = true` and `autoUpdateInput = false` (so no apply btn)
However, after clicking date will do nothing, need to add... 
>$("#yourinput").on('hide.daterangepicker',` function(ev, picker) {
> $(this).val(picker.startDate.format("YYYY/MM/DD"));
> });
(just provide for others who also encounter the problem)

dangrossman:
This is Example 5 on the website already

http://www.daterangepicker.com/#ex5

a55660988:
Okay, thank you for your prompt reply!!

PR #1171, opened 2016-07-18
Fix configuration builder for ranges not working with stringified dates

peetc1:
When testing using the configuration builder, I was getting an invalid date when a date range was added using the configuration builder. This was due to the format that was being used "MM/DD/YYYY" against a ISO formatted date does not work. Therefore, I changed it to test and see if the format would work, and if not just try and create a moment object without the format.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -82,6 +82,15 @@
         this.isShowing = false;
         this.leftCalendar = {};
         this.rightCalendar = {};
+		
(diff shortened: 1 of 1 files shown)


dangrossman:
Any use of `moment(string)` will generate warnings because this is deprecated functionality in the moment library. These warnings will generate endless issues in my repository complaining about them. That's why I always parse strings using a format string only.

PR #1173, opened 2016-07-23
UI clean up to show today, tomorrow and yesterday when date range fit…

kenkogi:
What do you think of this Dan? 
If the date range is all day today, tomorrow or yesterday, then the input does not need to have
[Jul, 22 2016 - Jul, 22 2016]. I thought it would be a good user experience for one of my projects.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -32,7 +32,13 @@
         this.parentEl = 'body';
         this.element = $(element);
         this.startDate = moment().startOf('day');
+        this.startOfToday = moment().startOf('day');
+        this.startOfYesterday = moment().subtract( 1, 'days' ).startOf('day');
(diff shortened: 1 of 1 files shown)


dangrossman:
I'm glad you found this useful, but I can't accept this pull request. I don't think there should be special case behavior for a few specific dates. This code doesn't handle the fact that the input is parsed and not just used to display dates. And I can't accept changes like this that change behavior people reply on; this library is deployed on tens of thousands of websites and those that expect their input to contain a date and not a word can unexpectedly break. Finally, you can make this exact behavior without changing any code in the library, by setting `autoUpdateInput: false` in your settings and putting this logic in your callback.

kenkogi:
Thanks Dan. I did not realize I could do that. I'll give it a try today.

Issue #1056, opened 2016-04-19
singleDatePicker without an initial date


parker789:
Hey Dan, 

I've looked around on here trying to find a solution to my problem, and have checked few different issues i thought were around the same idea, specifically #815 . I have the datepicker set using the code below, and all works.. but upon initialization the opened date picker displays the current date as already selected, so i can't select the current date.

```
$('#datepicker1').daterangepicker({
  singleDatePicker: true,
  autoUpdateInput: false,
}, function(date) {
  $('#datepicker1').val(date.format());
});
```
![screen shot 2016-04-19 at 8 53 59 am](https://cloud.githubusercontent.com/assets/4635313/14639777/52960e8c-060c-11e6-963d-a1c2103852d9.png)

dangrossman:
This library does not support "no selected date". There is always a selected date. If you don't specify one via `startDate`, it will default to today's date. If you want a single date picker without a selection, you will want to choose a different library.

parker789:
Thanks for the speedy reply, I was beginning to think that was the case.

rmm5t:
@dangrossman Might a reasonable solution just be to allow the `callback` to be called if the selected date is the same as the currently selected date (well, that combined with `autoUpdateInput: false`)?

Right now, the `callback` is ignored if the same date as `startDate` is selected.

rmm5t:
For future onlookers, here's another solution to this problem (use the `apply.daterangepicker` event to workaround the current known limitations of singleDatePicker usage):
https://github.com/dangrossman/bootstrap-daterangepicker/issues/815#issuecomment-220874445

BohdanZhmud:
There is also a case when user manually inputs some incorrect value. Currently single datepicker applies previous date, but with empty intial value example provided in documentation it does not work this way. You may use 'hide.daterangepicker' to change value of input control using picker.startDate to workaround this limitation.

yanup2913:
mmmmmmmmmmmmmmmmmmmm

Issue #815, opened 2015-09-24
singleDatePicker  - How to have initial empty date?


uniteddevelopers:
I have searched a lot but found no solution. It would be kind if anyone can tell me, how to have an initially empty date Input field with the single Date Picker. 

On the one hand this shows to my users that they can (but most not) enter a date. secondly the "non set date" is a valid value itself. it says "unlimited" but when a user enters a date this is a "limited until the date given". So both are completely different and valid/required values and my app is required to have both. But I struggled completely how to get this implemented. I found nothing, even after along search. Thank you very much.

dangrossman:
Set the `autoUpdateInput` option to false.

uniteddevelopers:
thanks! wow, that was a speed reply. I will try and afterswards write a little tutorial here, how to do it, so others can benefit and use this feature too.

dangrossman:
It's all in the one-page documentation.

superdummy:
@dangrossman thanks. it would be great if you can further "split" this flag up into two states: One for the initial Input field (to be initially not updated/touched) and then a second flag for updates,when a daterange was selected by the user with the daterangepicker (as this logic should not be disabled, but is currently).

@all
My findings:  `autoUpdateInput` is not really usable for having an initially empty datefield yet: it indeed allows to have initially empty fields (great), but it also disables the daterangpicker functionality: when the flag is set, a user picking a date will not be reflected/set in the input field. and this makes the datepicker completely useless.

superdummy:
Update:

for all interested (and it would still be very great if dan would improve the datepicker, as it is a really cool datepicker), the code is very clean and you can simply hack the lines below be removing the `&& this.autoUpdateInput` at the two places below:  Then you have an initially empty field but when the user selects a date it will be reflected in the input field. 

`
  updateElement: function() {
            if (this.element.is('input') && !this.singleDatePicker && this.autoUpdateInput) {
                this.element.val(this.startDate.format(this.locale.format) + this.locale.separator + this.endDate.format(this.locale.format));
                this.element.trigger('change');
            } else if (this.element.is('input') && this.autoUpdateInput) {
                this.element.val(this.startDate.format(this.locale.format));
                this.element.trigger('change');
            }
        },
`

dangrossman:
No no no... you don't need to hack the code. Just provide a callback function, and it'll be called every time a date is chosen. Update the input in that function.

```javascript
$('yourinput').daterangepicker({
[...]

Issue #788, opened 2015-08-28
autoUpdateInput issue


Toccata6767:
I'm trying to implement a single date picker.

With autoUpdateInput set to true (default behavior), the textbox will always have a date. You can change the date with manual entry or via the calendar and it automatically populates the textbox. There is no way to make it a blank value. By default it loads the current date.

With autoUpdateInput set to false, the textbox can be blank, but selecting a date from the calendar no longer works. Only manual entry will put a date into the field.

I don't think that is the intended behavior is it? I think the page should be able to load with a blank value and then allow entry either manually or from the calendar, unless I'm mistaken.

dangrossman:
This is the intended behavior. When you set `autoUpdateInput` to false, your input will not be automatically updated. You have to update it yourself, which you can do in a callback function. Your callback is invoked any time the selected date changes. 

Examples:
http://www.daterangepicker.com/#usage
http://www.daterangepicker.com/#ex3

PR #1176, opened 2016-07-27
Example for single date picker with empty value

gauravmak:
To address the issues like #788 #815 #1056

### DIFF
--- a/website/index.html
+++ b/website/index.html
@@ -289,6 +289,34 @@
                             });
                             </script>
 
+                            <p>
+                                If you do not wish to keep default value for the single date picker and your 
+                                field is optional, You may set the <code>autoUpdateInput</code> to false and
+                                use the callback function to set the value of the input when user selects date.
+                            </p>
+
+                            <div class="row">
+                                <div class="col-md-9 col-xs-12">
+                                    <script src="https://gist.github.com/gauravmak/c26b276bf03e969dc40beaef0e6f9b77.js"></script>
+                                </div>
+                                <div class="col-md-3 col-xs-12">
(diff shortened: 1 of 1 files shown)


dangrossman:
There is already an example of using the date picker with an empty value

http://www.daterangepicker.com/#ex5

gauravmak:
But, that example if for date ranges only and not for single date picker. apply and clear buttons are not available in case of single date picker, right?

andysh888:
*** I'm not good developer and have not enough experience with git. ***

Not fine solution, But looks like solution. Needs to modify code of daterangepicker.js
Modifications:
after line
`this.alwaysShowCalendars = false;`
add line
`this.emptyInitUpdate = false;`

after block
```
        // update day names order to firstDay
        if (this.locale.firstDay != 0) {
            var iterator = this.locale.firstDay;
            while (iterator > 0) {
                this.locale.daysOfWeek.push(this.locale.daysOfWeek.shift());
                iterator--;
            }
        }
```
add block
```
        if (typeof options.emptyInitUpdate === 'boolean') {
            this.emptyInitUpdate = options.emptyInitUpdate;
        }
```

in updateElement change line
`            } else if (this.element.is('input') && this.autoUpdateInput) {
`
to line
`            } else if (this.element.is('input') && (this.autoUpdateInput || (this.singleDatePicker && this.emptyInitUpdate))) {
`

Init datepicker sample:
```
	$('.daterange-single').daterangepicker({
		singleDatePicker: true,
		autoUpdateInput: false,
		emptyInitUpdate: true,
	});
```

PR #1186, opened 2016-08-04
Add ability to customize the icon prefix.

gillygize:
For our project, we needed the ability to customize the prefix that is used in front of the font icon.  This pull request allows the font awesome or glyphicon prefix to be customized.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -50,6 +50,8 @@
         this.autoUpdateInput = true;
         this.alwaysShowCalendars = false;
         this.ranges = {};
+        this.fontAwesomePrefix = "fa";
+        this.glyphiconPrefix = "glyphicon";
 
         this.opens = 'right';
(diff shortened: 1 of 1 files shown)


qed2000:
Might this not work better by setting the font awesome icons as a default and then allow the user to replace the fa class with their own icon class name? That could make it easier to use with other icon libraries in the future and prevent breakages if either library changes names or formats going forward.

gillygize:
Probably, if the development team wants to go forward that way.

thepauleh:
+1 @qed2000 - really liked the sound of this as we're using icomoon, an 'iconPrefix' would be more suitable in our use case, then we can drop glyphicons from our bundler

dangrossman:
My recommendation would be a private repo/fork that has the icon class you need, this one will continue assuming Bootstrap since it's a Bootstrap component

PR #1187, opened 2016-08-04
Added beforeDisplay callback for greater flexibility

drcrane:
I needed to set the dates from a custom chart component when the button was clicked, this seemed to be the best way and should allow complete flexibility. For example usage my callback looks like this:

    beforeDisplay: function(self) {
        if (chart.hasRange()) {
            self.setStartDate(moment.utc(chart.getStartTimeInMillis()));
            self.setEndDate(moment.utc(chart.getEndTimeInMillis()));
        }
    }

You can also stop the picker being displayed if the current state of your application means that a date range cannot be selected.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -77,6 +77,7 @@
         };
 
         this.callback = function() { };
+        this.beforeDisplay = function(self) { return true; };
 
         //some state information
         this.isShowing = false;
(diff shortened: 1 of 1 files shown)


dangrossman:
Could you better explain why you need this event? Why can't you set the dates on page load, or when your chart component loads, or whatever?

drcrane:
The chart changes with user interaction and so values would be stale after the page/chart load. Of course there are multiple ways of doing this, I could add a callback to the zoom/pan function of the chart that updates the calendar but to my mind these are roughly equivalent. Finally, though I realise this is not much of a concern, there is little, if any, need to update the calendar _every time_ the chart display range changes.

Issue #1191, opened 2016-08-10
Callback is not being called when autoUpdateInput is false and setting values through setStartDate & setEndDate


avcs06:
Where ever the input is being updated automatically when autoUpdateInput is true, callback should be executed there if autoUpdateInput is false. Except in the initial scenario where empty input is needed.

dangrossman:
I think this is unnecessarily complicated, and the change in behavior would not be backwards compatible. 

If you're calling `setStartDate` or `setEndDate` in your code, you can trigger your callback yourself in the same code at the same time if that's behavior you want.

PR #1192, opened 2016-08-10
Callback is not being called when autoUpdateInput is false and setting values through setStartDate & setEndDate

avcs06:
Where ever the input is being updated automatically when autoUpdateInput is true, callback should be executed there if autoUpdateInput is false. Except in the initial scenario where empty input is needed.


avcs06:
#1191

Issue #886, opened 2015-12-02
Open ended ranges


EhsanKia:
Is there currently no way to choose an open ended range? As set either the start or end date as "null"? Imagine wanting to select all events up to a date, or all events after a certain date.

dangrossman:
There is not a way to choose an open ended range. I would suggest if you're predefining ranges, using dates far in the past or the future to indicate all time. I'm sure you don't have events scheduled in 2099.

EhsanKia:
I was thinking something like Google search's date range, where you can pick a start, end, or both: ![](http://i.imgur.com/KbrUY2p.png)

I guess I'll just do it with two single date pickers, thanks.

sday:
I would also find use for a configuration allowing optional start or end dates.

EoinGriffin-AI:
+1 I've been asked to implement this feature at work for the 2nd time (the 1st time I looked into it and then told them no because of how hacky the solution would need to be, but they're persistent).

The suggested solution from @dangrossman isn't all that helpful ("_You might also consider changing your interface, because choosing "all events after a date" only requires picking one date, not a date range. You only need a single date picker, not a date range picker._"). In the same view the user might want a range to be open or closed.
Unless I can think of a better solution, I'll have to make some clunky switch to toggle between an open-ended and closed date range, and have a way to get them to choose whether it's the start of end date.

PR #1209, opened 2016-08-19
openEndedRange allows for null end date #886

nutso:


nutso:
#886

dangrossman:
This is not something I want to support

PR #1223, opened 2016-08-27
compactElementDisplay Option

lookatmike:
Adds a `compactElementDisplay` option; when enabled, and when the start and end of the selected range are the same, it will cause the element to display as a single date (and/or time) string instead of a range. So you'd see something like `1/1/2000` instead of `1/1/2000 - 1/1/2000`.

This differs from `singleDatePicker` in that you're still able to select a multi-day range if you want to; it just eliminates the redundancy in the display if you happen to pick start==end.

It's just looking at the final formatted start/end strings and displaying one if they're the same, so it works with times as well, although admittedly not in any clever way.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -49,6 +49,7 @@
         this.linkedCalendars = true;
         this.autoUpdateInput = true;
         this.alwaysShowCalendars = false;
+        this.compactElementDisplay = false;
         this.ranges = {};
 
         this.opens = 'right';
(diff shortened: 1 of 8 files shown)


dangrossman:
You can set `autoUpdateInput: false` and put two lines of code into your callback function to accomplish the same thing without any new options. Therefore, I do not wish to maintain all this extra code.

````
if (start == end)
  $(this).val(start.format('MM/DD/YYYY'));
````

lookatmike:
Yes, that's what I've been using on a few different projects, it seemed
like a common enough use case that it'd be nice to have natively.

That said, it is of course your project, and I certainly get not wanting to
add even more to the options list. Thanks anyway for the consideration!

Best,
Mike

PR #1235, opened 2016-09-12
cleaned up js

qed2000:
This Pull Request fixes many linting issues, such as: variable hoisting, improper use of double versus triple equals, enforced curly braces for conditional blocks. There are many optimizations that can be made, but this PR primarily covers linting issues.


dangrossman:
Thanks but no thanks. This does all kinds of things that make it more complicated for a human to read and understand.

PR #1241, opened 2016-09-18
Fix parent element. `parentEl` must be parent.

iegik:
After obfuscation `this.parentEl` looses its real variable.


dangrossman:
I don't want to change "body" to "this.element.parent()" as the default parent element. That's a breaking change.

PR #1245, opened 2016-09-20
[Feat](Options): Add inline option

TooHTooH:


WebberTakken:
Looks good, please pull

Liiva:
The background color is off when the inline picker is placed in an element with a non-white background; for example a bootstrap well.

dangrossman:
Hey guys. Thanks for the PR. Just wanted to let you know I'm still here, and hope to have time this week to review all the outstanding pull requests and start merging things in.

TooHTooH:
Hey @Liiva , I added the background color only when it's a tooltips but i can't reproduce the other issue that you tell me about the focus, can you tell me on which browser you are ? Or if you can send me a fiddle or anything ? Thx for your feedback

Liiva:
@TooHTooH, I added a jsfiddle here: https://jsfiddle.net/rdvxaaqb/
The layout is a bit weird (not sure why) but when you click on the top input field and press tab the inline picker disappears.

TooHTooH:
Oh ok, I think I get it, I'm going to patch it, I was setting up the daterangepicker on a `div` element instead of an `input`, I'm gonna check the Blur event ASAP 
Thx

TooHTooH:
@Liiva , Ok I think that's better : https://jsfiddle.net/b1epcrxu/

dangrossman:
It looks messed up because the daterangepicker.css stylesheet wasn't loading

It also doesn't work very well with options that affect layout...

https://jsfiddle.net/a9sp1qhk/1/

gabrielalan:
No thoughts on this? I really need this option :(

TheBetpet:
Any chance this could be checked and implemented? It looks so messy with the calendars overflowing onto the next row and it's a nightmare to try and override the css to make it work.

HansSchouten:
In case someone still needs an inline daterangepicker, I made a solution that does not require to change the original code: https://jsfiddle.net/9m63neo2/

PR #1254, opened 2016-10-02
Fix parent element. `parentEl` must be parent

iegik:


dangrossman:
There's no technical reason it needs to be the immediate parent by default, while this introduces risk of breaking existing websites where the picker is currently attached to the body. If the element the picker is attached to is like a panel or modal or something else of fixed size, the picker may be cut off and be partially invisible. As such, I will decline this PR to maintain backwards compatibility, but thank you.

iegik:
There was two fixes:
1. Fix position when more then one datepicker on page
2. Major fix - default parent element noir root element.

I can revert second change, to leave bug fix only, not changing default behavor!?

PR #1258, opened 2016-10-04
Implemented the updateInputOnDayHover and updateInputOnRangeHover options

luigonsec:
it has been implemented two new options to allow the user programatically control the input updates when he hover the days on the calendar or the defined ranges.


BKB503:
Thanks luigonsec this works like a charm now, no more confusion

dangrossman:
I do not think this needs to be an option, there has not been any widespread request for such a thing or complaints about the default behavior the past 5 years

PR #1273, opened 2016-10-15
Fix: the end of the range cannot before the start

vancepym:
If the end of the range is before the start of the range, should not display the range at all.


dangrossman:
Don't pass impossible ranges in your options

PR #1275, opened 2016-10-18
Add singleCalendar option

cmeyer90:
Adds option to select multiple dates from a single calendar view by enabling singleCalendar option.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -38,6 +38,7 @@
         this.dateLimit = false;
         this.autoApply = false;
         this.singleDatePicker = false;
+        this.singleCalendar = false;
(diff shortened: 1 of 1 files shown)


sytskehaagsma:
@cmeyer90 First off great option :) but it only worked for me when i added this to daterangepicker.js after the line 234. Or else the option is always false in a weird way :).

` if (typeof options.singleCalendar === 'boolean') {
            this.singleCalendar = options.singleCalendar;
   }`

dangrossman:
Interesting idea, but this widget has enough options to support IMO. If anyone asks about such a feature in the future I'll point to your repo.

PR #1289, opened 2016-11-04
Upgraded Date Picking Behaviour

JorinRBM:
This pull request expands the way you can select dates inside of the daterangepicker. 

Changelist:
- There is now only one state for the date picker: You have one or more selected.
- You are no longer explicitly selecting beginning and end dates.

New date selection flows:
- Day A selected && Later Day B selected: Range A-B
- Day A selected && Earlier Day B selected: Range: B-A
- Day A && Day B selected && you select outside of range: range extended
- Day A && Day B selected && you select endpoint A or endpoint B: single day selected
- Day A && Day B selected && you select between A and B: single day selected

This flow seems very natural to me and doesn't stop anyone from using the daterangepicker the same way it was originally designed.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -1289,78 +1289,100 @@
 
         },
 
-        clickDate: function(e) {
-
-            if (!$(e.target).hasClass('available')) return;
-
-            var title = $(e.target).attr('data-title');
-            var row = title.substr(1, 1);
-            var col = title.substr(3, 1);
-            var cal = $(e.target).parents('.calendar');
-            var date = cal.hasClass('left') ? this.leftCalendar.calendar[row][col] : this.rightCalendar.calendar[row][col];
-
-            //
-            // this function needs to do a few things:
-            // * alternate between selecting a start and end date for the range,
-            // * if the time picker is enabled, apply the hour/minute/second from the select boxes to the clicked date
-            // * if autoapply is enabled, and an end date was chosen, apply the selection
-            // * if single date picker mode, and time picker isn't enabled, apply the selection immediately
-            // * if one of the inputs above the calendars was focused, cancel that manual input
-            //
-
-            if (this.endDate || date.isBefore(this.startDate, 'day')) { //picking start
-                if (this.timePicker) {
-                    var hour = parseInt(this.container.find('.left .hourselect').val(), 10);
-                    if (!this.timePicker24Hour) {
-                        var ampm = this.container.find('.left .ampmselect').val();
-                        if (ampm === 'PM' && hour < 12)
-                            hour += 12;
-                        if (ampm === 'AM' && hour === 12)
-                            hour = 0;
-                    }
-                    var minute = parseInt(this.container.find('.left .minuteselect').val(), 10);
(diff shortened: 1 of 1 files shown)


l0c0luke:
this is close to what i was hoping for but my comments are that by selecting only a single date when you pick inbetween A & B, is annoying. sometimes i just want to move it by 1 day to start but not affect the end date.

JorinRBM:
I agree that being able to pick dates inside of a range and having the desired end move to that date would be amazing! But how do you know what end to move into the range? For example, Google analytics "solves" this by forcing you to be in one of two states; picking the beginning or picking the end. But in my opinion this is a higher cost to pay than having to reselect your endpoints if you want to bring the range in. A possible solution is to add in two checkboxes that enable you to specify that you are selecting either beginning or end, but the UI clutter would be ugly. If you can think of another way to track state, then I would be stoked to hear. Its a tricky little problem.

l0c0luke:
In my case, i think google analytics does it exactly like i am being asked to. Which is, i am always trying to create a range. So the highlighted input field is what is going to be set relative to the already existing range.

if you already have a & b... and a is highlighted... as long as you click range somewhere before b, the range updates. if you click after b, then a and b are set to equal.

same logic follows for setting end date.

its unlikely i will be shifting an already existing range. i will be editing one or the other. i think this is what google analytics does well.

JorinRBM:
If you have a different use case but like where the daterangepicker is at in this PR, then changing its behaviour is isolated to a single function. You could easily rewrite this version to accomplish what you are hoping for by creating a single boolean value that tracks which end of the range that you are changing, and setting endpoint A to the date clicked or to endpoint B if it is past A (earlier or later accordingly). In this way you can recreate the google analytics datepicker.

jgeurts:
@dangrossman: Any chance of getting this merged in? We recently upgraded from v1 and the v2 usability is throwing off our users.

dangrossman:
This PR doesn't return to the v1 date picking behavior. Why don't you "downgrade" back to v1 which your users prefer?

jgeurts:
I would prefer to stay current with the latest version, if possible and I think the behavior described by @JorinRBM should cover things from a usability standpoint. Are you open to having options for date selecting behavior? If so, is there anything that can be done with this PR (or a separate) to help get it merged?

daveknapik:
Hello! I just found this PR after reviewing the list of open issues and PRs to see if someone had done what I wanted and... I want this! Can we breath new life into it? I'm willing to pitch in and help, just let please us know what would make this more mergeable, @dangrossman. Thanks!!

JorinRBM:
Thank you so much @daveknapik! Hopefully others can pick it out and use it too!

dangrossman:
I appreciate this PR, and played around with it a lot, and asked some people for feedback, and it wasn't clearly better than the way it currently works. A lot of people found choosing a different range newly confusing now that what would happen was unpredictable. Sometimes the existing range was being extended, sometimes not, it wasn't clear how to get back into a state where they could choose the specific dates they wanted. So I'm passing on this now.

PR #1302, opened 2016-11-18
Implemented 'init.daterangepicker' event dispatched when the calendar…

lightenbody:
Added a new event 'init.daterangepicker' dispatched when first time rendered the calendar. I am not sure if this should be also in renderTimePicker function.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -50,6 +50,7 @@
         this.autoUpdateInput = true;
         this.alwaysShowCalendars = false;
         this.ranges = {};
+		this.initialized = false;
 
         this.opens = 'right';
         if (this.element.hasClass('pull-right'))
@@ -620,7 +621,10 @@
         },
 
         renderCalendar: function(side) {
-
+            if (!this.initialized) {
+                this.element.trigger('init.daterangepicker', this);
+                this.initialized = true;
+            }
             //
             // Build the matrix of dates that will populate the calendar
             //


dangrossman:
What purpose does this event serve?

lightenbody:
To set the start values during the rendering of the calendar.
I have created the following

$(_element).daterangepicker
        singleDatePicker: true
        showDropdowns: true
        showISOWeekNumbers: true
        locale:
          firstDay: 1
        ,
        # on click callback
        (start, end, label)->
          startDate = moment(start).startOf('isoWeek').add(1,'days')
          endDate = moment(start).endOf('isoWeek')

          if _onChangeCallback then _onChangeCallback(startDate, endDate)

      $(_element).on 'apply.daterangepicker', (event, picker)->
        picker.setStartDate moment(picker.startDate).startOf('isoWeek')
        picker.setEndDate moment(picker.startDate).endOf('isoWeek')

      $(_element).on 'init.daterangepicker', (event, picker)->
        picker.setStartDate moment().startOf('isoWeek')
        picker.setEndDate moment().endOf('isoWeek')

Unfortunately, startDate and endDate does not work when singleDatePicker is set to true, so that is why i had to create init event.

I want to be able to select a period in singleDatePicker because even if user select 2016-11-18, it must give me back the selected range of first day of week and last.

Do you have then any other idea to get this?

PR #1313, opened 2016-11-29
Customize font classes for icons

carlsaunderhds:
Implement logic to customize the icons used to render the calendars. See examples and documentation on the new options:-

- calendarClasses
- clockClasses
- leftArrowClasses
- rightArrowClasses


dangrossman:
Thanks, but I think this library has enough options in general. If someone must change the icons, they can do so locally or in a fork.

PR #1315, opened 2016-12-06
disable user select when selecting a date range

Bilal-io:
It was annoying to see the text highlight when selecting a date range, so I got rid of that using css.

### DIFF
--- a/daterangepicker.css
+++ b/daterangepicker.css
@@ -59,7 +59,8 @@
   .daterangepicker.dropdown-menu {
     max-width: none;
     z-index: 3001; }
-  .daterangepicker.single .ranges, .daterangepicker.single .calendar {
+  .daterangepicker.single .ranges,
+  .daterangepicker.single .calendar {
     float: none; }
   .daterangepicker.show-calendar .calendar {
     display: block; }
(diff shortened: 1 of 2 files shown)


dangrossman:
What exactly does this do? The text in the text box is user editable, people should be able to interact with it like any other text box.

Bilal-io:
@dangrossman my bad there is a stupid plugin in my text editor that formatted the files and it looks like I have changed a lot of lines. I will redo the whole thing.
the only thing I added was:
`-moz-user-select: none;     /** All Firefox **/
 -webkit-user-select: none;  /** All Chrome and Safari **/
 -ms-user-select: none;      /** IE 10+ **/
 user-select: none;          /** Future implementations **/`
and it has nothing to do with the text box. it has to do with the numbers in the calender, days 1 to 30. when you click on a day and drag to select a range, it highlights the text. where it shouldn't for a better look. compare the two images. 
1-before-change
2-after-change

dangrossman:
You don't drag to select a range

Bilal-io:
I guess you're right, I had to test that again to see that my initial thought was wrong. I thought I could drag to select before, but it isn't true. Sorry about that.

PR #1318, opened 2016-12-12
Adding labels and contrast to input fields

Uetzicle:
This update adds options for increasing contrast on the input fields when choosing From and To dates. Adds clarity to what each side is, and which is being acted on.


dangrossman:
This PR is too difficult to review, so I'm going to opt not to.

PR #1330, opened 2016-12-24
Added window as alternative to currently-scoped instance of window

achang28:
This slight addition is meant for solutions that do not create an instance of the window object.  When imported within a client-side .js file (e.g. during file bundling), the "this" variable that is passed into the IIF (line 28 of daterangepicker.js) is an empty object.  As I understand it, "this" should be a separate instance of window when daterangepicker.js is sourced from within an html file.

My apologies if there isn't something I have correctly understood.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -25,7 +25,7 @@
         // Browser globals
         root.daterangepicker = factory(root.moment, root.jQuery);
     }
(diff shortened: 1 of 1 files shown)


dangrossman:
I'm afraid I don't know how to verify or test this, so I'm going to stick with following the Universal Module Definition template.

Issue #1332, opened 2016-12-27
Click on month label to select full month


SamanthaAdrichem:
I'd like a feature where if you click on the month label it will select that month as your date range

dangrossman:
Thanks for the PR. Since there hasn't been much request for a feature like this, I am opting not to support it.

PR #1333, opened 2016-12-27
#1332

SamanthaAdrichem:
When you click on a month label it will select that month as the date range


dangrossman:
Thanks for your PR. Since there hasn't been much request for a feature like this, I am opting not to support it.

PR #1350, opened 2017-01-13
remove restriction to type of input for setting the initial value

Jerome2606:
I use MVC Razor engine to generate my html and Datetime field input is of type datetime. I was wondering that daterangepicker can handle this kind of input if I set up the value as expected.
I modify just one line to remove the restriction of having a input of type text.
Only check if the element is an input element


dangrossman:
The HTML5.1 specification for a datetime element places strict rules on acceptable values, which this library is not coded to enforce. I am therefore passing on this PR.

PR #1352, opened 2017-01-13
Add hidden class to days that are outside of the current month

periclesmacedo:
While editing the css to customize the datepicker, I see that, while hovering, the the days of other months in the current calendar can't get their css overriden without overriding the ones from the next or previous months that are also disabled. This change should allow the current user to update the css of those specific days without overriding all the disabled days of the pre/next calendar days.

### DIFF
--- a/daterangepicker.css
+++ b/daterangepicker.css
@@ -115,6 +115,8 @@
     background-color: #357ebd;
     border-color: transparent;
     color: #fff; }
+  .daterangepicker td.hidden {
+    visibility: hidden; }
   .daterangepicker th.month {
     width: auto; }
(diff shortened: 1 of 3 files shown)


dangrossman:
I try not to make changes that would break live websites; it should be safe for someone to update this library and only get bug fixes and new features, not breaking changes. This is a breaking change.

periclesmacedo:
Oh, ok, I see how that can be a breaking change now. How about if I just add a class without any extra css rules? So the only changes would be something like:
```
799 -                        classes.push('off');
799 +                        classes.push('off', 'out-of-month');
```

Issue #1361, opened 2017-01-20
[Feature] Add methods to set min/max date


lukasamd:
As in title, it is useful for custom data.

PR #1362, opened 2017-01-20
Add set min/max date methods (issue #1361)

lukasamd:
Add set min/max date methods (issue #1361)

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -491,6 +491,22 @@
             this.updateMonthsInView();
         },
 
+        setMinDate: function(minDate) {
+            if (typeof minDate === 'string')
+                this.minDate = moment(minDate, this.locale.format);
+
+            if (typeof startDate === 'object')
+                this.minDate = moment(minDate);
+        },
+
+        setMaxDate: function(maxDate) {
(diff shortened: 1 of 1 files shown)


dangrossman:
Thank you for the suggestion, but I do not think it's necessary to have/support special methods to change a single option. The minDate/maxDate and any other options can be easily changed the same way they were initially set.

morrow95:
I realize this is an old thread, but can you explain how one would change the minDate and maxDate after it has already been initialized?

I've tried something like the following, but can't seem to get it working : 

```
$('#drp-el').data('daterangepicker').minDate = 'new moment in correct format I specified when originally initializing'
```

I've been using this for years now, but never had a situation where I needed to modify the min/max date on demand after it was already initialized.  Basically, based on other things that occur, I want to change these so the next time the picker is opened the new min/max are respected.

dangrossman:
Initialize it again with the new settings

morrow95:
@dangrossman  👍  Works just as I hoped... I just put my init code in a function then can call it as needed.  The values I need to change in the init are dynamic so nothing else needed - just works.  Thanks - happy holidays and new year!

PR #1364, opened 2017-01-21
Disable option for input hover behavior

danpetrikin:
Honestly thought it was a bug. Saw the comment on the site that you were not intending to add this support, which is fine. I've gone ahead and added it for my needs. You're welcome to use the changes or toss them.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -170,7 +170,7 @@
             }
         }
         this.container.addClass(this.locale.direction);
-
+		
         if (typeof options.startDate === 'string')
             this.startDate = moment(options.startDate, this.locale.format);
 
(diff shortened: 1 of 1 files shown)


thecristen:
I'm using this and I just wanted to thank you for adding this! The hover changing the displayed dates was really confusing my users so it's nice to be able to easily turn that off now. :)

marktisham:
Second that here. We had several clients confused by the default hover behavior. The behavior in this PR works a lot more intuitively for the newbies, thanks for creating! +1 vote for merging into the master!

prashantpawar:
PR #1411 also deals with the same issue. It's obvious that a lot of people are facing the same issue (including us). Unfortunately, none of these PRs change this behavior as an 'option' like `disableInputHoverOnSelection` or something like that so that the people relying on the existing behavior don't get affected.

PR #1371, opened 2017-01-24
minYear rightCalendar equals currentYear leftCalendar;

DMITRYTISHKIN:
before right calendar has min year equals current year


TCruise:
Thanks dude You solved my problem too. Hats off to you

yanikore:
Not really useful as a lot of our customers think the year is unchangeable because they haven't selected a starting date in the left side.

PR #1387, opened 2017-02-12
Option added to define a custom icon system

RobsyRocket:
Hi,

i had the need to modify the library because the project i working on uses a different icon system.

The implementation now provides the option to use a own icon system instead of the required Font Awesome and/or Glyphicon. This was not possible before, except maybe with Mutual Observers, due the hardcoded CSS definitions and recreation of the whole calendar HTML structure after each paging.

The icon system option accepts a string or array to support one ore multiple icon systems – like the existing implementation does it hardcoded with Font Awesome and Glyphicon. A new prototype function in Daterangepicker, named createIconClass, does the prefixing and joining of icon classes.

Please review my changed and i would appreciate it if they get merged into the main repo.

Thank you already
Robert

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -29,7 +29,7 @@
     var DateRangePicker = function(element, options, cb) {
 
         //default settings for options
-        this.iconSystem = 'glyphicon';
+        this.iconSystem = ['fa', 'glyphicon'];
         this.parentEl = 'body';
         this.element = $(element);
         this.startDate = moment().startOf('day');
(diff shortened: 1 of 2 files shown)


dangrossman:
Thank you for the PR, but I am not merging it for two reasons:

1. This library is specifically advertised as a Bootstrap component, so it supports the Bootstrap icons

2. This PR does not do what it purports, "supporting custom icon systems", since it just changes the prefix; a different icon system might have entirely different icon class names too, not just a different prefix

RobsyRocket:
I'm fine with your complains agains my request. But, i have a question before i put any effort in another implementation and pull request. 

Would you a consider a modified implementation which triggers additional events, especially for the paging of the calendars or the renderCalendar function itself? This would add the possibility to listen for changes without using MutualObservers, for whatever use case the user has.

dangrossman:
Nobody else has asked for events when the calendars are paged in the past 5 years, so I would not add them. I don't maintain code only used by a single person -- that belongs in your copy not this one. Do not bother with the PR.

PR #1394, opened 2017-02-17
Hide datepicker when hitting ESC key

Stafie:
- Close datepicker when hitting ESC

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -440,7 +440,7 @@
             });
         } else {
             this.element.on('click.daterangepicker', $.proxy(this.toggle, this));
-            this.element.on('click.daterangepicker', $.proxy(this.toggle, this));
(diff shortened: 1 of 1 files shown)


daterangepicker.js
```diff
@@ -440,6 +440,7 @@
             });
         } else {
             this.element.on('click.daterangepicker', $.proxy(this.toggle, this));
+            this.element.on('click.daterangepicker', $.proxy(this.toggle, this));
```
urkle: @Stafie why is this duplicated?

urkle: @Stafie shouldn't this be
```javascript
this.element.on('keydown.daterangepicker', $.proxy(this.keydown, this));
```

Stafie: @urkle Yeah, my mistake. Pushed a fix for that.

Issue #1386, opened 2017-02-10
Custom Range date selection issue


skasgarv:
**Issue 1:**
When choosing a date range, assume Jan 1st to Jan 31st, first date field takes Jan 1st and second date field is highlighted. After clicking the second date, **the first date field is re-highlighted and date changes from Jan 1st to Jan 31st** until the mouse is on the calendar. When we move the mouse away from the calendar, then we can see the spread from Jan 1st to Jan 31st.

**Issue 2:** 
A second, even more-confusing case is where you want to edit just the second date of the range. Again, in the case of Jan 1st to Jan 31st, imagine you’d like to edit the second date to Jan 14. When I click on the “Jan 31, 2017” field to select it to be changed, **the first date field is re-selected and repopulated with the new date which should have the second date field**. This could obviously cause confusion, sending some users into an endless cycle of clicking dates. This appears to be a bug with the component.

Please look into the issue and I hope it can be resolved. Thanks

dangrossman:
Issue 1: This is not an issue. The inputs reflect what your mouse is hovering over to preview what clicking will do.

Issue 2: I could not reproduce at http://www.daterangepicker.com

skasgarv:
@dangrossman I am able to replicate Issue 2 at  http://www.daterangepicker.com 
I double checked. The first field is getting re-populated. 
Please check the attached image
![dateissue](https://cloud.githubusercontent.com/assets/18016940/22846202/627ee048-efb4-11e6-83c0-303fc8dfd3fb.png)

dangrossman:
Your reproduction steps said "When I click on the “Jan 31, 2017” field to select it to be changed", but I do not see your cursor in the second text box

skasgarv:
The image attached is after clicking Jan 31st field and changing it Jan 15th. So the spread should be from Jan 1st to Jan 15th but in the image you can see that its Jan 15th to Jan 15th. The cursor moves back to the first date field as soon as I click a date.

dangrossman:
When I click Jan 31st and change it to Jan 15th, I see Jan 1 to Jan 15 in the text boxes. Are you still hovering your mouse over Jan 15 and that's why you see 1/15 to 1/15? That's "issue 1" again , which is not an issue.

skasgarv:
Hey. Sorry for the delayed response. I was travelling.
I did fork your repository and made some minor tweaks to suit my application.
[...]

PR #1411, opened 2017-03-06
Do not update range on hover.

tomlynch:
Users find the change on hover to be confusing. Because they have already made a selection, they think the dates are changing from the hover and they do not realize clicking apply will use the selected dates. Removing these lines allows updating the dates only when selected by the user.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -1260,12 +1260,6 @@
             var cal = $(e.target).parents('.calendar');
             var date = cal.hasClass('left') ? this.leftCalendar.calendar[row][col] : this.rightCalendar.calendar[row][col];
 
-            if (this.endDate && !this.container.find('input[name=daterangepicker_start]').is(":focus")) {
-                this.container.find('input[name=daterangepicker_start]').val(date.format(this.locale.format));
-            } else if (!this.endDate && !this.container.find('input[name=daterangepicker_end]').is(":focus")) {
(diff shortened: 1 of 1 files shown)


prashantpawar:
@tomlynch Another issue was filed related to the same problem and the author rejected the request as claiming that the behavior is intended. https://github.com/dangrossman/bootstrap-daterangepicker/issues/1386 

I'd suggest that we introduce a new flag which by default keeps the behavior as is, but when being turned on, disables the range on hover. This way the library does not change behavior unexpectedly for all the users who are using it.

programcsharp:
I've got a lot of users also being confused by the dates changing on hover and thinking they chose the wrong dates.

This is especially pernicious if you don't realize you can click twice in the calendar to select a range and instead click the start date input, then a date, and the end date input, and another date. It looks like it's a straight up bug that reset your dates.

I came here to suggest exactly what this PR does -- remove the input value updates on hover but leave the rest of the hover. I'd argue that the input updates on hover is to the point of a bug because it's so confusing unless you're very tech savvy.

I can't think of a use case that'd be broken by that -- are there some existing cases that require a hover value update? If there are, I could do up a PR with an option to re-enable the hover value update if necessary, but I just can't think of a possible need.

If anything, hover date value should be a different display than the from/to dates, or a tooltip, something separate from the from/to date boxes. If there is a need to display the current value of the date that you can see right there because you're hovering it, that'd be the way to do it...

PR #1414, opened 2017-03-08
Live Updating Date Inputs

lucasklaassen:
Adds ability to enable live updating when typing in the start and end date input ranges.

## Documentation:
1. You can now enable live calendar updating by setting `liveUpdating` to true in your date range options.

## Mergeability concerns:
1. If users are not using 4 character years (YYYY), the formatting will break when live updating is turned on. (Right now I have validation that prevents live updating when there is an invalid year being entered but this only applies to 4 character years (YYYY). See lines 1516 to 1525 in daterangepicker.js

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -35,6 +35,7 @@
         this.endDate = moment().endOf('day');
         this.minDate = false;
         this.maxDate = false;
+        this.liveUpdating = false;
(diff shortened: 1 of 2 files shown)


lucasklaassen:
@dangrossman any issues you see with merging this?

Issue #1185, opened 2016-08-04
Can't set minDate/maxDate in a created daterangepicker


dogcalas:
I create a daterange picker with this options.
```
    $('.daterange').daterangepicker({
        format: 'YYYY/MM/DD HH:mm:ss',
        timePickerIncrement: 5,
        timePicker12Hour: true,
        timePickerSeconds: false
    });

```
Then, I tried to change the minDate using this:
```
    $('.daterange').daterangepicker({
maxDate:'2016/01/01',
minDate:'2016/01/05'
    });
```
But the daterange not change.

dangrossman:
When you call `.daterangepicker` a second time, you are creating an entirely new date range picker. You need to pass all of the options, not just something you want to change. The default date format string, which you're not overriding in the second instantiation, is not YYYY/MM/DD, so those `minDate` and `maxDate` options are going to be seen as invalid as well.

dangrossman:
Also, `format` is part of the locale object, not a top-level option. Use something like this:

    $('.daterange').daterangepicker({
      locale: {
        format: 'YYYY/MM/DD HH:mm:ss'
      },
      timePicker: true,
      timePickerIncrement: 5,
      timePicker12Hour: true,
      timePickerSeconds: false,
      minDate: '2016/01/01 00:00:00',
      maxDate: '2016/01/05 23:59:55'
    }

StanleyNguma:
I had this problem and I was able to go around it.
On the daterangepicker.js file I added this function:

`setMinDate: function(minDate) {
            if (typeof minDate === 'string')
                this.minDate = moment(minDate, this.locale.format);

            if (typeof minDate === 'object')
                this.minDate = moment(minDate);

            if (!this.timePicker)
                this.minDate = this.minDate.startOf('day');

            if (this.timePicker && this.timePickerIncrement)
                this.minDate.minute(Math.round(this.minDate.minute() / this.timePickerIncrement) * this.timePickerIncrement);

            if (!this.isShowing)
                this.updateElement();

            this.updateMonthsInView();
        },`

Then on my HTML page I used the following line to call that function
` $('input[name="date_picker"]').data('daterangepicker').setMinDate(moment());`

I hope this helps

pjaaar:
Hi,

I just wanna let you know that I had the same needs as @dogcalas and the solution provided by @StanleyNguma is working

cornelRaiu:
@StanleyNguma you should also add 

[...]

PR #1416, opened 2017-03-10
add setMinDate() and setMaxDate() methods

cornelRaiu:
This seems to be functionality I am using on all projects where I use this date picker.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -83,6 +83,9 @@
         this.leftCalendar = {};
         this.rightCalendar = {};
 
+        //other options
+        this.updateFromValue = 'start'; //possible options 'start', 'end'
+
         //custom options from user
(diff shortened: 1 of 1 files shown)


dangrossman:
You can set new options whenever needed by calling `$.daterangepicker(options)` again. Additional functions are not needed. I'm not 100% sure what the `updateFromValue` thing is, but it seems to replicate the already-existing `autoUpdateInput` option.

cornelRaiu:
Not sure why the updateFromValue commit went in the pull request. That was only for my use. The pull request was for setMinDate and setMaxDate methods

proxy-m:
maybe use code from https://github.com/dangrossman/bootstrap-daterangepicker/issues/1185 ?

PR #1418, opened 2017-03-12
Include moment.js only if it does not already exist in window.

BradyLiles:
Some applications may have an odd-ball configuration set up with SystemJS or whatever there module loader could be and already have moment.js included and this prevents it from loading twice in those configurations.
This repeats the practice we're already doing with loading JQuery.


BradyLiles:
@dangrossman Is there any chance I can get an eta on a next patch release?

dangrossman:
Some time before the end of 2017?

PR #1423, opened 2017-03-17
Apply the date only when it's selected

gnunes76:
It was added a new configuration called "autoApplyOnHide" with default "true". If the user chooses to set this property to "false", the behavior of the component will be to update the field only when a date is selected.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -38,6 +38,7 @@
         this.maxDate = false;
         this.dateLimit = false;
         this.autoApply = false;
+        this.autoApplyOnHide = true;
         this.singleDatePicker = false;
         this.showDropdowns = false;
         this.showWeekNumbers = false;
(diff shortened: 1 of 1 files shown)


dangrossman:
What is this for?

eglyecyriaco:
Dan,
The problem we had was that every time we closed the singledatepicker it would set a date. 
We couldn't simply open and close the calendar without setting it to today's date, nor could have it empty once a date has been selected.

Basically, we still wanted to be able to use a singledatepicker with autoApply, but wanted to be able to open and close the calendar without having it picking a date for us, and wanted a way to clean the field, i.e. set it to null.

Giving that our field is nullable, and will likely never have a value (it's the end date for a message validity period), we found this solution would be easier to implement.

If there were a simpler way to accomplish this without this change, we would love to know about it.

dangrossman:
This can be accomplished with `autoUpdateInput: false` and updating the input in your callback function or `apply.daterangepicker` event listener. A new option isn't needed. Thank you though.

PR #1448, opened 2017-04-11
Add data-date to the td elements

turigeza:
I would like to have the date in some sort of format on the td elements in the picker so I can use the date on events like mouseover.

People in the past solved this in complicated way like the link shows.
http://stackoverflow.com/questions/35906930/bootstrap-date-range-picker-get-current-date-the-mouse-is-hovering-over

Any thoughts?


dangrossman:
This is not a feature requested by others in the past, so I will opt to pass on merging this into the main library. You should maintain a fork for your organization with your needed customizations.

Issue #1410, opened 2017-03-06
setEndDate() does not update time -- only date.


slgriffiths:
If you try and set the end date of the daterangepicker with `setEndDate()`, the date correctly updates, but the time (if using `timePicker` option) does not.

`setStartDate()` works for both date and time.

Rambuto:
+1 Big problem in project. Waiting for merge this commit to master branch

Kolano:
There seem to be similar issues with just initializing a datepicker where time components of both startDate and endDate are ignored.

Kolano:
There seem to be similar issues with just initializing a datepicker where time components of both startDate and endDate are ignored.

PR #1460, opened 2017-04-28
fix: setEndDate() does not update time -- only date #1410

yconoclast:


Ledragon:
+1 on this

raztus:
The commit that introduced this issue is 0e449c228 which fixed #1295. I have tested and verified that this fix, #1460, does not cause a regression for the issue reported in #1295, and should be merged, @dangrossman.

dangrossman:
Merged manually

PR #1483, opened 2017-05-23
Added an option to not dismiss the date-range picker when clicking outside of it

Erowlin:
- New option appeared in the initializer, so we can now specify whether we want the date picker to be dismissed when clicking outside of it. 
It avoids hacks like that: 
```coffeescript
$("#my-daterange-picker").on 'show.daterangepicker', (e) ->
    # Trick to unbind the dismiss of the datepicker when clicking outside it
    $(document).off 'mousedown.daterangepicker'
    $(document).off 'touchend.daterangepicker'
    $(document).off 'click.daterangepicker'
    $(document).off 'focusin.daterangepicker'
```

- Updated the documentation to reflect that new option
- Updated the generator to be able to use that option in it
- added the `outsideClick.daterangepicker` event in the documentation, which was first introduced by #489 .

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -51,6 +51,7 @@
         this.linkedCalendars = true;
         this.autoUpdateInput = true;
         this.alwaysShowCalendars = false;
+        this.dismissOnOutsideClick = true;
         this.ranges = {};
 
         this.opens = 'right';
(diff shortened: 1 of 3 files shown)


Erowlin:
Is this gonna be considered one day? :)

dangrossman:
Nobody else has asked for this behavior in 5 years, so I do not wish to maintain it as an option, it should be in your own copy only, thanks

adg29:
This option is of interest to me without having to maintain it in my own copy. Thanks.

PR #1487, opened 2017-05-27
Hook on auto apply

f00se:


dangrossman:
Two lines below that it calls `clickApply` which triggers `apply.daterangepicker`. In other words, you already have an event to listen to.

PR #1491, opened 2017-05-31
Datepicker not updating when changing the input value

imktks:
When a date like "4 May 2017" is selected from the datepicker and the user attempts to update manually the input to "4 May 2018", the datepicker doesn't respond.

There is a function that handles the keyup events on the corresponding input and performs validations before updating the datepicker value.

```
if (!this.element.is('input')) return;
if (!this.element.val().length) return;
if (this.element.val().length < this.locale.format.length) return;
```

In the third case, it checks the length of the datepicker's selected format (in our case "D MMMM YYYY" the length is 11) against the length of the input value (i.e. "4 May 2018", with length 10). As a result, the datepicker considers the new value as invalid.

I think, a more appropriate approach is to use moment.isValid function in order to check the validity of the current value.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -1565,7 +1565,7 @@
         elementChanged: function() {
             if (!this.element.is('input')) return;
             if (!this.element.val().length) return;
-            if (this.element.val().length < this.locale.format.length) return;
+            if (!moment(this.element.val().trim(), this.locale.format, true).isValid()) return;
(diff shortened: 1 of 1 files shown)


dangrossman:
Unfortunately this only ever works for single date pickers, since ranges like "4 May 2018 - 4 May 2018" are not valid dates.

PR #1492, opened 2017-06-01
Add localization option for range's labels.

feki:
Hi 

I add support for localize range's labels by using `opts.locale.ranges` option. 

### Motivation

In my project I need to store value of range in relative form ('Today', etc.), but the site supports multiple languages. So It would be nice to use labels as keys of ranges in relative forms, not as localized labels for users. Therefore I add new option for range's labels localization, see Usage below.

## Usage:

```javascript
$('input[name="daterange"]').daterangepicker(
{
  // ...
  ranges: {
    'today': [moment(), moment()],
    'yesterday': [moment().subtract(1, 'days'), moment().subtract(1, 'days')],
    'last7Days': [moment().subtract(6, 'days'), moment()],
    'last30Days': [moment().subtract(29, 'days'), moment()],
    'thisMonth': [moment().startOf('month'), moment().endOf('month')],
    'lastMonth': [moment().subtract(1, 'month').startOf('month'), moment().subtract(1, 'month').endOf('month')]
  },
  locale: {
    // ...
    ranges: {
      'today': 'Today',
      'yesterday': 'Yesterday',
      'last7Days': 'Last 7 Days',
      'last30Days': 'Last 30 Days',
      'thisMonth': 'This Month',
      'lastMonth': 'Last Month'
    }
  }
}, 
function(start, end, label) { });
```

If user doesn't provide ranges locale option, original labels are used (library works as before, https://github.com/feki/bootstrap-daterangepicker/blob/feature/add-localization-of-ranges-labels/daterangepicker.js#L353).

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -75,7 +75,8 @@
             customRangeLabel: 'Custom Range',
             daysOfWeek: moment.weekdaysMin(),
             monthNames: moment.monthsShort(),
-            firstDay: moment.localeData().firstDayOfWeek()
+            firstDay: moment.localeData().firstDayOfWeek(),
+            ranges: {}
         };
 
         this.callback = function() { };
@@ -169,6 +170,9 @@
                 var rangeHtml = elem.value;
                 this.locale.customRangeLabel = rangeHtml;
             }
+
+            if (typeof options.locale.ranges === 'object')
+                this.locale.ranges = options.locale.ranges;
         }
         this.container.addClass(this.locale.direction);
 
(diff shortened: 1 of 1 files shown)


dangrossman:
Range labels can be localized without any additional options

feki:
@dangrossman thank you to your quick reply. 

I know, that range labels can be localized without any additional options and up to now I use it in this way. 

But as I wrote above, my needs is about unambiguous identification of selected range. So the localized labels is not suitable for this purpose. So this PR is not only about localization of range labels. For that I create issue #1497.

dangrossman:
This is not something any other user has asked for in 5 years, so I will still opt not to merge and maintain it as part of this library. Code only you are using should be in your organization's copy/fork. Thanks.

PR #1494, opened 2017-06-01
Fix for outerWidth and outerHeight returning selector element

chopraanmol1:
outerWidth and outerHeight usage:

	https://bugs.jquery.com/ticket/12491


dangrossman:
That's a 5 year old ticket about a conflict with a specific version of jQuery UI. No changes are needed here.

PR #1523, opened 2017-07-04
Ensure that pressing the 'enter' key in the daterange inputs leads to consistent behaviour across major browsers

cliffsun91:
When you enter a date in the start/end date inputs and you press the 'enter' key, the calendar would then update with the range specified in the inputs. This works in Chrome/Firefox/Edge but not in IE 11. This change is to make this behaviour consistent across these major browsers.
As to the cause of this issue, I think it's down to the different browsers triggering the onChange event for inputs on different things. It seems that Chrome/Firefox/Edge will trigger the event when the 'enter' key is pressed as well as when the focus is lost on the input. Whereas IE 11 would only trigger the onChange event when focus is lost on the input. 

I know that this may be a bit specific of a change for IE 11 (which admittedly is because we need to support it for our clients), but this change means that we no longer need to know (or care) which 
browsers trigger the onChange event on the 'enter' key press.


cliffsun91:
Hi, can I get an update to whether this will be merged in and released? If not we'll have to fork it ourselves.

Issue #1527, opened 2017-07-06
Hide calendar inputs


toncid:
Hello,

Would it be possible to add a config option to toggle calendar inputs:

![screen shot 2017-07-06 at 12 47 57](https://user-images.githubusercontent.com/1837897/27907840-8730dc22-6249-11e7-85d1-a24af76bd018.png)

I cannot seem to find a way to do it except via external CSS. 

Suggestions? Thoughts?

toncid:
Added a pull request.

toncid:
Added a pull request #1528.

negativefix:
+1.  Having an option to hide inputs would be great.

stefkes:
+1 don't really the point of having a manual input on top of the calendar. Default hidden with option to show would make more sense.

arulnithi:
+1 I am currently using external CSS to hide this

PR #1528, opened 2017-07-06
Add showCalendarInputs option

toncid:
Add a setting that makes it possible to hide calendar inputs (shown by default).

Issue #1527.


dangrossman:
This can be accomplished with CSS/JS and hasn't been much requested

PR #1533, opened 2017-07-17
Add options to specify icon classes in the plugin

lalitindoria:
This change helps use different icons apart from font-awesome or bootstrap fonts. The new options have been added to the demo HTML as well.


dangrossman:
There are no longer any font icons

PR #1550, opened 2017-08-14
Add updateTextOnHover option

guywithnose:
The way the text box updates on hover can create a confusing user experience.  After selecting the end date, the start date text box changes to the end date because that is where the mouse is hovering.  This feels unexpected because an interaction with the end date seems to be changing the start date.


dangrossman:
No longer relevant

PR #1554, opened 2017-08-17
remove default body offset.

xzedx:
in some condition the <body> offset is not 0, and then the default offset would cause some issue


dangrossman:
Taking this change caused the picker to be rendered quite far away from the input it was attached to

PR #1557, opened 2017-08-23
Alternative mode for selecting date ranges

janssonav:
Added a new configuration flag (now named `endpointCalendars`) that changes the default date range selection behavior as follows:
- The left calendar always selects (or moves) the start date and hovering over the left calendar is reflected in the start date input (unless the input is focused).
- The right calendar does the same for the end date.

We need this new mode because some users consider the default behavior confusing and have requested this new functionality. Obviously, we don't want to break the default functionality which is why the new mode is behind a configuration flag that needs to be explicitly enabled.

The new mode doesn't play well with the `linkedCalendars` flag but I think it makes little sense to use the two flags together.

This pull request also eliminates some repetitive code.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -52,6 +52,7 @@
         this.autoUpdateInput = true;
         this.alwaysShowCalendars = false;
         this.ranges = {};
+        this.endpointCalendars = false;
 
         this.opens = 'right';
         if (this.element.hasClass('pull-right'))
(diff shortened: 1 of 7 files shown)


meraj-ahmed:
Thanks for adding this much-needed feature. I hope it gets merged!

KowsalyaPK:
These are necessary changes, as the user gets confused with the default behaviour, adding these will avoid those confusions. @dangrossman Can you please merge this PR?

dangrossman:
This was the behavior of V1 of this library, which you can still check out if you prefer. Head to the "Releases" tab to find v1.23 or older. It will not be supported in newer versions, as it is inconsistent with every other date range picker I can find in the wild.

PR #1558, opened 2017-08-24
Emit an event when a date (day) is selected

PortikNorbert:
This will make listening to the click possible.
Benefit: in a handler it will be possible to check if the start or end date has been selected for a range.


dangrossman:
You should not be taking any action when someone is in the middle of making a selection (that they may easily never complete or cancel). This event would be dangerous and misused.

PR #1561, opened 2017-08-31
New variables in outsideClick-function: this.lastclickStartDate / End

firstinstance:
Can be use with: 
jQuery("#daterangepicker").on('outsideClick.daterangepicker', function (ev, picker) {
	alert(picker.lastclickStartDate.format('YYYY-MM-DD'));
	alert(picker.lastclickStartDate.format('YYYY-MM-DD'));
	if (picker.lastclickStartDate && !picker.lastclickEndDate)
	{
		// do something
	}
});

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -1168,6 +1168,8 @@
                 target.closest(this.container).length ||
                 target.closest('.calendar-table').length
                 ) return;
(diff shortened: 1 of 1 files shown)


dangrossman:
You can keep track of old date selections in your own variables if you so choose, the library doesn't need to

PR #1585, opened 2017-10-17
Add usability: SingleDatePicker with Buttons

BrunoEngComp:
Usage SingleDatePicker with Buttons Apply and Cancel, without having to activate the timePicker property.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -383,7 +383,7 @@
             this.container.find('.daterangepicker_input input, .daterangepicker_input > i').hide();
             if (this.timePicker) {
                 this.container.find('.ranges ul').hide();
-            } else {
+            } else if(this.autoApply) {
                 this.container.find('.ranges').hide();
             }
         }


dangrossman:
Honest question: in that configuration, what purpose do the buttons serve? 

Once a date is clicked, the calendar has served its function, there's no reason for it to stay open on the screen waiting for "apply" to be clicked. If your entire form should wait for things to be chosen then "apply" should be part of your form, not the calendar. 

Similarly, clicking outside the calendar hides it, which is how people generally dismiss a calendar widget... there's no button to dismiss the calendar that pops up from the Windows taskbar for the past 16 years. If it's serving some kind of "clear" functionality, that's also something that could be part of your form or your input, rather than having to open a calendar to clear another UI element.

BrunoEngComp:
With this possibility, I can use the component events, such as:

User selects a date, but if he does not want to send more the date he selected? He does not have a button to clear the field.

In this case, we would need the active buttons to be able to use the event and clear the field.

So we could take advantage of the maximum of the component and would not disrupt its operation itself, since it would only be an activation flag.

Issue #1587, opened 2017-10-18
Fail to choose the same day in a custom range date when the day is the first day of daylight saving time


diegolido:
I realized this  'cause 15 oct 2017 was the first day of daylight saving time in Brazil and my app fails there.
I found the problem is because an issue with endOf('day') moment function (https://github.com/moment/moment/issues/3132), that still open there.
So I figured out how to deal with it by myself work around with the logic in datarangepicker.js.

Instead of:

        setEndDate: function(endDate) {
			debugger;
            if (typeof endDate === 'string')
                this.endDate = moment(endDate, this.locale.format);

            if (typeof endDate === 'object')
                this.endDate = moment(endDate);

            if (!this.timePicker)
                this.endDate = this.endDate.endOf('day');

            if (this.timePicker && this.timePickerIncrement)
                this.endDate.minute(Math.round(this.endDate.minute() / this.timePickerIncrement) * this.timePickerIncrement);

            if (this.endDate.isBefore(this.startDate))
                this.endDate = this.startDate.clone();

            if (this.maxDate && this.endDate.isAfter(this.maxDate))
                this.endDate = this.maxDate.clone();

            if (this.dateLimit && this.startDate.clone().add(this.dateLimit).isBefore(this.endDate))
                this.endDate = this.startDate.clone().add(this.dateLimit);

            this.previousRightTime = this.endDate.clone();

            if (!this.isShowing)
                this.updateElement();

            this.updateMonthsInView();
        }

Use:

        setEndDate: function(endDate) {
			debugger;
            if (typeof endDate === 'string')
                this.endDate = moment(endDate, this.locale.format);

            if (typeof endDate === 'object')
                this.endDate = moment(endDate);

            if (!this.timePicker)
                this.endDate = this.endDate.add(1,'d').startOf('day').subtract(1,'second');

            if (this.timePicker && this.timePickerIncrement)
                this.endDate.minute(Math.round(this.endDate.minute() / this.timePickerIncrement) * this.timePickerIncrement);

            if (this.endDate.isBefore(this.startDate))
                this.endDate = this.startDate.clone();

            if (this.maxDate && this.endDate.isAfter(this.maxDate))
                this.endDate = this.maxDate.clone();

            if (this.dateLimit && this.startDate.clone().add(this.dateLimit).isBefore(this.endDate))
                this.endDate = this.startDate.clone().add(this.dateLimit);

            this.previousRightTime = this.endDate.clone();

[...]

PR #1588, opened 2017-10-18
Fix for issue #1587

diegolido:
I realized this 'cause 15 oct 2017 was the first day of daylight saving time in Brazil and my app fails there.
I found the problem is because an issue with endOf('day') moment function (moment/moment#3132), that still open there.
So I figured out how to deal with it by myself work around with the logic in datarangepicker.js.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -501,7 +501,7 @@
                 this.endDate = moment(endDate);
 
             if (!this.timePicker)
-                this.endDate = this.endDate.endOf('day');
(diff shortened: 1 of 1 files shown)


Horlacher:
looks like a workaround to me

diegolido:
Show us the right way please.

dangrossman:
A workaround is better than nothing, thank you

phazei:
workaround only fixes a single day while breaking the prior day, so doesn't actually fix this

screen shot 2018-12-26 at 11 18 51 am

PR #1596, opened 2017-10-30
Add 2 more options for more flexible

phattranky:
Hi,

I would like to add 2 options:

disableInput: sometimes I wouldn't like user input in 2 fields because they may type unexpected characters
emitChangeOnSameDate: It's allow plugin call selectedDate also in case the user choose same date

I think 2 cases is need for all. Please help me merge it. Thanks

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -52,6 +52,7 @@
         this.autoUpdateInput = true;
         this.alwaysShowCalendars = false;
         this.disableInput = false;
+        this.emitChangeOnSameDate = false;
         this.ranges = {};
 
         this.opens = 'right';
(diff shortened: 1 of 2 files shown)


phattranky:
Hi @dangrossman, Could you please help me check this PR. Thanks

phattranky:
@fidalgo @MaikuMori 
could anyone help me merge this PR. Thanks

Horlacher:
**disableInput:** You should always expect invalid characters as this widget is executed on the client. You can still disable the input fields with css:

.daterangepicker_input input {
    pointer-events:none;
    color:#aaa;
    background:#f5f5f5;
}

dangrossman:
You can disable or hide inputs with CSS, and you can be notified when same date is chosen again by listening for the `apply.daterangepicker` event.

PR #1598, opened 2017-10-31
Keep time in ranges without timePicker

mansona:
I needed a way to keep the idea of having times associated with ranges but I didn't want to display the time picker. 

I think I have covered all of the cases I need to, let me know if I need to update the PR.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -44,6 +44,7 @@
         this.showWeekNumbers = false;
         this.showISOWeekNumbers = false;
         this.showCustomRangeLabel = true;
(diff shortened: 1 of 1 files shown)


dangrossman:
This seems odd and nobody else has asked for this behavior. I would recommend instead if you want to display or store a static time component for each date, you do so in your callback function or elsewhere in your own code.

mansona:
I don't think it's unreasonable to alter the code to "not clobber the times I gave it", from my perspective it's odd that it would make any changes to the parameters that were passed in the first place 🤔

PR #1620, opened 2017-12-05
Highlight Days param and momentTimezone

sbeluzo:
Setting a New param to Highlight days and set moment Timezone compatibility if timezone available

e.g.
`highLight : ["05/12/2017","06/12/2017"]`


dangrossman:
Hi there,

There is already an option for highlighting days called `isCustomDate`, and use of timezone code in this library has only ever lead to unsolvable bugs all over the place.

Kind regards,
Dan

PR #1657, opened 2018-01-22
Added option to allow an empty value

vtrottier:
I am using this library for a filter input, so I propose to add an option to be able to have an empty value.


dangrossman:
You should use `autoUpdateInput: false` to accomplish this. A new option is not needed.

vtrottier:
Correct me if I'm wrong, but it seems to me that doing so I lose the date validation. Is there currently a way to keep it while also allowing empty value?

PR #1658, opened 2018-01-23
Update daterangepicker.js

ajaxboy:
.show()  equivalent to   display: block.   firefox displays calendars (with date range) on a single column one on top of the other, while chrome does it on two columns (one next to the other). ' display: flex' fixes the issue and make both browsers display one next to the other.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -1122,7 +1122,7 @@
             this.previousRightTime = this.endDate.clone();
 
             this.updateView();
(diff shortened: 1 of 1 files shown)


dangrossman:
But there's no such issue to fix...

ss

ajaxboy:
I guess it might be because of the specific scenario on the custom format of the page being worked on.  Ahh! maybe because I am using bootstrap 4.

PR #1687, opened 2018-03-08
Comparison dropdown

PandelisZ:
Implementing autorange selection as a range


PandelisZ:
OOPS! Didnt mean to make this here!

Issue #1580, opened 2017-10-13
Feature request: an event after the first date is selected + option to advance the calendar


amarnathm:
Great software!

Feature request: an event after the first date is selected will be helpful in some situations. Also, while adding text to each date is possible, will be helpful to make this process easier. Finally, an option to change the opened months after the first date is selected will be helpful. 

Thanks!

dangrossman:
Hi, there's nobody here that takes and develops feature requests, but thanks

shailu26:
i also want this feature cause for custom range my requirement is when use selectes any start date from custom range than it should able to select only 30 days

PR #1700, opened 2018-03-22
triggering an event when start date is selected

MihaitaN:
It also handles #1580


dangrossman:
I do not want people to be messing with things when only a start date has been clicked. The UI is redrawn so often that it just wastes peoples' time.

Issue #1702, opened 2018-03-24
When using bootstrap 4, setting autoApply=true doesn't hide the buttons.


ravithb:
Due to the 'hide' class not being available in bootstrap 4, the autoApply functionality doesn't hide the buttons when set to true.

dangrossman:
Fixed via other means

PR #1703, opened 2018-03-24
Fix for issue  #1702, where the 'Apply' and 'Cancel' buttons where st…

ravithb:
…ill visible when 'autoApply' was set to true and bootstrap version 4 was being used.


dangrossman:
Thanks, this bug no longer exists

PR #1736, opened 2018-05-07
use the `:text` selector instead of `input[type=text]`

thorn0:
See http://api.jquery.com/text-selector/


dangrossman:
Sounds like a good change, thanks

thorn0:
Yep. It took me a long time to understand why things don't work. Turned out my input didn't have the type attribute.

PR #1749, opened 2018-05-16
Bug with minYear & maxYear without min/max Date

TNAJanssen:
Fix for maxDate.month is not a function and minDate.month is not a function


Alonski:
Hey, just wondering what the is the status of this PR

PR #1757, opened 2018-05-24
make custom label selected after click

aiddroid:
make custom label selected after click.
related issue: 
 [Custom Range should be selected in expanded panel after click "Custom Range" ](https://github.com/dangrossman/daterangepicker/issues/1756)

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -1163,6 +1163,8 @@
             var label = e.target.getAttribute('data-range-key');
             this.chosenLabel = label;
             if (label == this.locale.customRangeLabel) {
+                this.container.find('.ranges li').removeClass('active');
+                this.container.find('.ranges li:last').addClass('active');
                 this.showCalendars();
             } else {
                 var dates = this.ranges[label];


JonHolt:
@dangrossman please merge this one.

daterangepicker.js
```diff
@@ -603,9 +603,11 @@
             this.renderCalendar('right');
 
             //highlight any predefined range matching the current start and end dates
-            this.container.find('.ranges li').removeClass('active');
-            if (this.endDate == null) return;
-
+            if (this.endDate == null) {
+                return;
```
aquamoth: Looks like two two rows are really just switched, but the commit confuses things by introducing the else-section, which makes it less obvious that `this.container.find()...` and `this.calculateChosenLabel()` go together.

aiddroid: Maybe addClass/removeClass "active" should move into calculateChosenLabel()

hasmanyguitars:
We've had a customer request this as well.

yubiuser:
@aiddroid 

Is there any progress on this?

PR #1782, opened 2018-06-21
ability to use single calendar with date ranges

csalwin:
Add option to allow single date range mainly for mobile use where space is limited. First stages but works for me i haven't tested with all the different options though, but would be a good feature to have.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -40,6 +40,7 @@
         this.maxSpan = false;
         this.autoApply = false;
         this.singleDatePicker = false;
+        this.singleRangePicker = false;
         this.showDropdowns = false;
         this.minYear = moment().subtract(100, 'year').format('YYYY');
         this.maxYear = moment().add(100, 'year').format('YYYY');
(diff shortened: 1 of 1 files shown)


lizardkhing:
I have a need to select multiple dates but noy necessarily in a range. For
example a sport team schedule. Closet ive seen allows it bit when u go to
edit or add another date previously selected get wiped out. Any ideas?

dangrossman:
@lizardkhing Your requirement is not related to this pull request. Selecting multiple dates on a calendar is not a feature of this library. You may, of course, take the code here as a starting point to programming a calendar that works that way for your website. I hope it helps you get a head start on that job.

braver:
I like this because of the stated application of providing a range input on smaller screens (i.e. mobile devices). One issue here is that in [updateFormInputs](https://github.com/dangrossman/daterangepicker/blob/b87331638c39d402d43bf5d4e089fb497b8ec292/daterangepicker.js#L1013) it should account for the `singleRangePicker` application. E.g.:

```js
if (this.singleDatePicker && ! this.singleRangePicker) ||   ... etc
```

instead of 

```js
if (this.singleDatePicker ||
```

@dangrossman are you interested in this feature? I could try and finish it.

Issue #1783, opened 2018-06-23
Is the `locale.direction` option officially supported?


stegano:
Is the `locale.direction` option officially supported? I found this feature in `demo.html` but it does not seem to work well.

If set `options.direction` to `ltr`, the calendar position of the previous month and the calendar position of the next month are reversed, as shown in the image below(this seems to be a bug).

![2018-06-23 10 42 46](https://user-images.githubusercontent.com/11916476/41810270-e19471b2-7736-11e8-8bf7-a7ad6c10e73a.png)

PR #1784, opened 2018-06-23
fix(#1783): fixed a css bug that occurrence when `locale.direction` is ` ltr`

stegano:
- fix(#1783): fixed a css bug that occurrence when `locale.direction` is ` ltr`

Now it works! 😃


stegano:
ping @dangrossman  😺

PR #1799, opened 2018-07-19
setMaxDate functionality

eusoikonomou:
Created setMaxDate method in the daterangepicker's constructor so that if someone uses live dates for the datepickers he/she can still update selected date values without having problem with the upper limit of the maxDate. This way you can ensure (if needed) that the user can choose moment() as upper limit even if time is not static.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -449,6 +449,22 @@
 
         constructor: DateRangePicker,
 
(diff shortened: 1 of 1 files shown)


dangrossman:
You can change options at any time by recreating the picker with the new options

PR #1818, opened 2018-08-19
Added disabled date feature

ar3s:
I've added the option to pass an array of dates that should be disabled on rendering.
it must be an array of dates in format 'YYYY-MM-DD' or an empty array.

Init example:

_var list = ['2018-08-28', '2018-09-06', '2018-10-31'];
$('#inputpicker').daterangepicker( {disabledDates: list} );_

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -54,6 +54,7 @@
         this.autoUpdateInput = true;
         this.alwaysShowCalendars = false;
         this.ranges = {};
+        this.disabledDates = [];
 
         this.opens = 'right';
         if (this.element.hasClass('pull-right'))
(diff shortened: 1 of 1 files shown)


dangrossman:
There's already a setting to disable dates, it's `isInvalidDate`

ar3s:
How does it works?

MohammadRezaAlizadeh000:
> isInvalidDate

like this:

$(function () {
                    opens: 'left',
                    isInvalidDate: function (date) {
                        return ['2022-09-10', '2022-09-15'].includes(date.format('YYYY-MM-DD'));
                    }
                }
                });
            });

PR #1814, opened 2018-08-14
Add option to allow selection of end date first

georgebohnisch:
Added a boolean option that allows the control to be initialized with the ability to select an end date first.

Example: User selects January 15 (start date is set) and then selects January 10 (normally this would reset the start date to January 10). If the option `allowEndDateFirst` is set to true, the start date is set to January 10 and the end date is set to January 15.

PR #1823, opened 2018-08-24
Adds a mode "leftStart", restore earlier keyboard-input dates (with flag "typingInput")

MPagel:
leftStart means you pick the start date from the left panel and you pick the end date from the right panel exclusively.

also functionize a few common code segments (highlightRange and hoverDefinedRange)


MPagel:
leftStart is similar in ways to #1814 and #1778

GregoireDucharme:
That would be pretty useful

PR #1836, opened 2018-09-10
Allow optional/empty date on initial load

ducqui:
$("input.datepicker").daterangepicker({emptyDate: true, singleDatePicker: true});

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -54,6 +54,7 @@
         this.autoUpdateInput = true;
         this.alwaysShowCalendars = false;
         this.ranges = {};
+        this.emptyDate = false; // meant to be used with singleDatePickerg
 
         this.opens = 'right';
         if (this.element.hasClass('pull-right'))
@@ -276,6 +277,9 @@
 
         if (typeof options.alwaysShowCalendars === 'boolean')
             this.alwaysShowCalendars = options.alwaysShowCalendars;
+
+        if (typeof options.emptyDate === 'boolean')
+            this.emptyDate = options.emptyDate; // allow date input to be empty on initial load
 
         // update day names order to firstDay
         if (this.locale.firstDay != 0) {
@@ -1112,7 +1116,7 @@
             }
 
             //if a new date range was selected, invoke the user callback function
-            if (!this.startDate.isSame(this.oldStartDate) || !this.endDate.isSame(this.oldEndDate))
+            if (!this.startDate.isSame(this.oldStartDate) || !this.endDate.isSame(this.oldEndDate) || this.emptyDate)
                 this.callback(this.startDate.clone(), this.endDate.clone(), this.chosenLabel);
 
             //if picker is attached to a text input, update it
@@ -1348,6 +1352,7 @@
         },
 
         clickApply: function(e) {
+            this.emptyDate = false; // date has been populated and can no longer be empty
             this.hide();
             this.element.trigger('apply.daterangepicker', this);
         },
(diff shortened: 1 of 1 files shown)


dangrossman:
You don't need to create a new options for this, "input initially empty" is the name of one of the examples on the documentation site already: http://www.daterangepicker.com/#example5

alexandrejobin:
I think we must separate the need to have a input with an empty value and the need to have the input by updated as soon as you click on a date.

It's true that if you use autoUpdateInput: false, you can have an empty input but i need to also auto update the input field when you click on a date.

How can i do that?

dangrossman:
@alexandrejobin The code for that is right here: http://www.daterangepicker.com/#example5

alexandrejobin:
I saw the example but i just thought that natively, the library should allow an empty input instead of manually do everything (update the value, format the value).

Since that we can't init the input with an empty value AND still get the autoUpdateInput, i come up with this solution to have a startDate and endDate controlled separatly.

```
function init()
{
    var startDate = $("#StartDate");
    var endDate = $("#EndDate");

    attachDatePicker(startDate, null, endDate.val());
    attachDatePicker(endDate, startDate.val(), null);

    startDate.on("change", function (e)
    {
        attachDatePicker(endDate, startDate.val(), null);
    });

    endDate.on("change", function (e)
    {
        attachDatePicker(startDate, null, endDate.val());
    });
}

function attachDatePicker($input, minDate, maxDate)
{
    $input.daterangepicker({
        singleDatePicker: true,
        autoUpdateInput: false,
        minDate: minDate,
        maxDate: maxDate
    }, function (start, end, label)
    {
        // update input
        this.element
            .val(start.format("YYYY-MM-DD"))
            .trigger("change");
    });
}
```

ducqui:
@dangrossman I realize that example#5 has "input initially empty". That's not the bug.
The bug is that you click the datepicker, want to choose Today (default start date), and NOTHING!!!
Why? The startDate and the selectedDate are the same and it WILL NOT trigger the update.

My code change was to allow initial input empty but also to update the date to Today on user selection.

The workaround is to choose another date, then choose Today again. That will work with the existing code, but that is BAD UX.

dangrossman:
@ducqui The example does not have that fault. Unlike the callback function, the `apply` event is triggered even if you select the already selected date range. The example code uses that event. Feel free to try it out at http://www.daterangepicker.com/#example5.

toxpal:
I know it's an old topic, but I discovered your plugin just yesterday, and it's amazing. 

The only confusing thing is that additional JS code is needed to initialize picker with empty value. In all other plugins, empty values are initialized by default, and you write extra code only if you want to add something (not to remove something).

I also strongly believe there should be a simple option to initialize picker with empty value (or even better - empty value should be used by default if input field is like `<input type="text" name="whatever" value="">` to avoid need of extra JS code, but that's just my opinion, and developer probably knows a good reason why this feature is not available.

shushenghong:
@dangrossman i think we need an extra option to keep init input with empty value, it's not the same meaning with autoUpdateInput, 
if user click to choose some thing, autoUpdateInput will auto update the ui

shushenghong:
@toxpal totally agree with u

shushenghong:
my suck solution
```
var oval = $picker.val();
$picker.daterangepicker(xxxx);
if(!oval) {
    $picker.val('');
}
```

aledmb:
> you click the datepicker, want to choose Today (default start date), and NOTHING!!!
> Why? The startDate and the selectedDate are the same and it WILL NOT trigger the update.

yeah, that's my issue too.

kunambi:
Sorry to necro this thread, but it's very strange that the PR wasn't allowed. 

The option described - which imho should be default - is very different from `autoUpdateInput: false`, which breaks `autoApply: true`.

PR #1883, opened 2018-12-05
Feature/single calendar

jeroen-groenveld:
With the singleDatePicker option you can only select 1 date (hence the name) and not a range between two dates. With the singleCalendar option only one calendar is shown, but you can still pick a range. This is useful for mobile, because when it shows the two calendars it's quite long.

This is the same pull request as #1873 instead of pulling from the master branch it's pulling from another branch.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -41,6 +41,7 @@
         this.maxSpan = false;
         this.autoApply = false;
         this.singleDatePicker = false;
+        this.singleCalendar = false;
         this.showDropdowns = false;
         this.minYear = moment().subtract(100, 'year').format('YYYY');
         this.maxYear = moment().add(100, 'year').format('YYYY');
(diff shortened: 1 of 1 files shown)


morganhvidt:
Cool! I hope this gets pulled in :)

phpMagpie:
+1 I really like this, ideally would be able to have this on mobile and the 2 calendar view on larger screen sizes.

I had an issue with line #390 applying the "auto-apply" class so changed line #389  from `if (!this.timePicker) {` to `if (this.singleDatePicker && !this.timePicker) {`

luckyexpert:
Yes I need this.. I need this calendar to be more mobile friendly, and with this option I can check width of the screen to determine if I should show 1 calendar or 2. 2 calendars just doesn't fit on mobile nicely..

asacarter:
I was looking for this feature. I've not tested yet but I hope this feature gets added to master.

koljada:
I also need this feature. Sometimes having 2 calendars is an overkill

PR #1886, opened 2018-12-13
Feature to dynamically limit the max days of range

carolinesalib:
Using this configuration, the user only can select a limited range of days after the selected start date.

When using this configuration, the cancel button is replaced with a reset button. It makes the exact same thing, except for closing the
calendar.


dangrossman:
There's already an option for limiting the range of days after the selected start date, `maxSpan`. The cancel/reset button thing seems unrelated and specific to your use case.

PR #1895, opened 2018-12-26
Revert "Fix for issue #1587"

phazei:
Reverts dangrossman/daterangepicker#1588

This fixes one day but breaks the day before, on top of breaking time consistency with all the other dates all as detailed out here:
https://github.com/dangrossman/daterangepicker/issues/1890#issuecomment-450014401

screen shot 2018-12-26 at 11 18 51 am

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -489,7 +489,7 @@
                 this.endDate = moment(endDate);
 
             if (!this.timePicker)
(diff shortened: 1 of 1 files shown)


phazei:
A true fix for this issue should be in the next release of moment.  Actually required a entire refactor of the start-end-of methods.  Rather non-trivial.

https://github.com/moment/moment/pull/4338#issuecomment-446857625

PR #1906, opened 2019-01-13
Allow for selection of week from Week label.

Scott-G:
Allows the whole week to be selected by clicking the Week Label


Scott-G (commented):
Also accepts 'dayoffset' as hours to allow allow the selected week to begin and end at a set time (eg 7am). Usefull for calculating working week or weekly production etc.

PR #1911, opened 2019-01-22
bug:2018-12-31为2018年01周

oyai32:
It is wrong to turn the last few days of the year into weeks，
For example,2018-12-31为2018年01周


dangrossman:
I'm not sure what the bug is or what the fix is supposed to do. Perhaps it belongs upstream in the moment.js library?

oyai32:
When choosing December 31, 2018, it should show the last week of 2018, or the first week of 2019. So showing the first week of 2018 is wrong

PR #1920, opened 2019-02-06
Clean up fortnight option settings

tadywankenobi:
Created the correct fallbacks so if showFortnightNumbers is set, without showWeekNumbers or showISOWeekNumbers being set, correct layout and values will be applied.

Add two new options:

* **showFortnightNumbers**
* **fullWidthTitle**

**showFortnightNumbers**

Setting adds a "fortnight" column to the left of the calendar with each second week given a number. Respects showWeekNumbers and showISOWeekNumbers options if set. Not dependant on them. Set to false by default.

**fullWidthTitle**

Setting makes the left (previous) chevron align fully to the left, regardless of whether the showWeekNumbers, showISOWeekNumbers or showFortnightNumbers options are set. Uses correct table colspan variable to respect width of Month/Year field on table. Set to false by default.

Not sure what packaging tools to use so only including daterangepicker.js and demo.html. Manually modified js for my purposes (included).


dangrossman:
I appreciate you sharing this code. However, nobody else has requested the ability to display fortnight numbers in 7 years, so it's not an option I want responsibility to maintain.

Issue #1872, opened 2018-11-16
Single datepicker hides cancel/apply buttons even when autoApply is false


andyfurniss4:
I want to have a single datepicker with the apply/cancel functionality but it seems that regardless of whether I set autoApply to false, the buttons are always hidden when singleDatePicker is true.

Solybum:
For some reason, if singleDatePicker is enabled but timePicker is not, it will add auto-apply as seen here
https://github.com/dangrossman/daterangepicker/blob/bc645d93e837905dcd38b04b8d9123978ea9f571/daterangepicker.js#L384-L392

dangrossman:
Because these buttons are not needed in that situation. There is no chance of a partial selection.

Solybum:
I was looking about having the input to be cleared, which if I understood correctly, requires 'autoUpdateInput': false, and 'autoApply': false, so the user can remove the content from the input and not have it be filled again automatically.

Solybum:
Actually, nvm, it seems you can instead just populate the field from the callback, and letting the user delete the input on their own, although for my use case I might want to use the cancel button as a clear instead.

jonathantribouharet:
Any news on this?
I need to have a simple way for users to clear the input, the clear button seems perfect.
Here my code:

```
    datepicker.daterangepicker({
      singleDatePicker: true,
      autoApply: false,
      autoUpdateInput: false,
      locale: {
        format: 'DD/MM/YYYY',
        cancelLabel: 'Clear'
      }
    });
```

dangrossman:
I don't expect to change this. I would suggest either changing it in your copy, or putting your clear button on your form, instead of on the date picker dropdown widget.

TBSliver:
I have added a PR which fixes this issue, #1931 (Ignore the earlier one, Github has decided to default to allow PR's on forks instead of back to the main repo and I didnt notice until to late).

If this change is not wanted, then the documentation should be updated to reflect that the singleDatePicker cannot be used with this option, to save confusion.

PR #1931, opened 2019-03-20
Allow for singleDatePicker to disable autoApply

TBSliver:
As found in #1872, the auto apply option is ignored when using a single date picker, meaning that autoApply is by default set. This change allows for the autoApply config option to be used even in a singleDatePicker.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -386,7 +386,7 @@
             this.container.find('.drp-calendar.left').addClass('single');
             this.container.find('.drp-calendar.left').show();
             this.container.find('.drp-calendar.right').hide();
-            if (!this.timePicker) {
+            if (!this.timePicker && this.autoApply) {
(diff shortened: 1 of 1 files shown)


lacostenycoder:
Any idea if and when the PR can be merged?  We would really appreciate this feature!

sebandres:
@dangrossman any particular reason why there are so many PRs that seem good to merge but aren't? I was thinking of contributing an option to update the calendar when the first date is selected but feels without purpose if the feature will never be merged :(

zvala:
> @dangrossman any particular reason why there are so many PRs that seem good to merge but aren't? I was thinking of contributing an option to update the calendar when the first date is selected but feels without purpose if the feature will never be merged :(

@dangrossman same here, I was thinking about contributing... Any insight??

Issue #1829, opened 2018-09-03
sass file removed


deefeloper:
Hi, between v3.0.1 and v3.0.2 the daterangepicker.scss was removed... why was this done?

I want to use this component in a project which compiles and combines sass files to a single css files, but it does not process the .css file and instead just puts `@import url(../../../../node_modules/daterangepicker/daterangepicker.css);` in the final css which offcourse doesnt work.

The reason I request this is because sass is deprecating loading normal css files as if they were sass files with the following announcement:
```
DEPRECATION WARNING on line 1, column 8 of /home/dave/projects/floraconnect-online-frontend/app/assets/styles/pages/index.scss:
Including .css files with @import is non-standard behaviour which will be removed in future versions of LibSass.
Use a custom importer to maintain this behaviour. Check your implementations documentation on how to create a custom importer.
```

dangrossman:
Someone contributed a sass verssion of the V2.x stylesheet as a one-time thing. I don't use sass myself, so I don't know how to write the files and I don't have the tooling for them. I wanted to make a V3.x version of this library. Thus no more sass file.

dangrossman:
No, probably not. I don't write sass, so why should there be a sass file I can't work on in the repo?

deefeloper:
it's compatible with css and gives much cleaner code than without, for example

the following sass:
```
a {
  color: #aaa;

  &hover {
    color: #bbb;
  }

  span {
    float: right;
  }
}
```

will translate to the following css:
```
a {
  color: #aaa;
}

a:hover {
  color: #bbb;
}

a span {
  float: right;
}
```

xmon:
I agree that it would be very helpful to have the styles in a .scss file in the repository.

Like @deefeloper I also compile all my .CSS in a single file, most repositories have the styles in .scss files.

webdevnerdstuff:
I just put in a pull request where I converted the CSS to SCSS for version 3.0.5. I haven't tested it yet, but it's there if you want it.

https://github.com/dangrossman/daterangepicker/pull/1951

PR #1936, opened 2019-04-03
Include datetimepicker.scss

xmon:
#1829


xmon:
renamed daterangepicker.scss

PR #1954, opened 2019-05-15
Add `select with in date sections` option

HechengLi:
Added select date within date sections, if setup like `selectInSection=[{ startDate: moment('2019-05-01'), endDate: moment('2019-05-10') }]`, then if you select start date as 2019-05-02, then the max end date you can select is 2019-05-10.

### DIFF
--- a/daterangepicker.css
+++ b/daterangepicker.css
@@ -196,6 +196,10 @@
   border-radius: 0;
 }
 
+.daterangepicker td.range-end {
+  border-right-color: #357ebd;
+}
+
 .daterangepicker td.start-date {
   border-radius: 4px 0 0 4px;
 }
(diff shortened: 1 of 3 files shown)


dangrossman:
There are already minDate and maxDate settings for this

HechengLi:
> There are already minDate and maxDate settings for this

what i meant is you can have multiple sections [{ startDate: moment('2019-05-01'), endDate: moment('2019-05-10') }, { startDate: moment('2019-05-11'), endDate: moment('2019-05-15') }], and you cant select start date from one section and end date in another, not exactly the same as minDate and maxDate.

dangrossman:
I don't quite understand why you'd want these "sections" or what use cases they would enable. Nobody has ever asked for such a feature in the 7 years since I wrote this library. So, I do not want to maintain this setting, as you would be the only person using it. Thanks for sharing the code anyway.

PR #1990, opened 2019-07-19
get current parentEl instead of element during initialization

zhernosekim:


dangrossman:
I don't know what this does or why you're doing it.

PR #2023, opened 2019-09-23
Do not hide calendar when alwaysShowCalendar is true via toggle

AdamHess:


dangrossman:
Clicking prev/next only updates the calendars, `toggle()` is not called at any point?

https://github.com/dangrossman/daterangepicker/blob/master/daterangepicker.js#L1214

AdamHess:
It's called of your previous and next buttons are outside of the date
picker box

dangrossman:
The only place in the entire codebase that the toggle function is called is in the click and keydown events for your page element the picker is attached to.

AdamHess:
Take a look at this example: 

image

        this.container.find('.drp-calendar')
            .on('click.daterangepicker', '.prev', $.proxy(this.clickPrev, this))
            .on('click.daterangepicker', '.next', $.proxy(this.clickNext, this))

and later on 
     
    this.element.on('click.daterangepicker', $.proxy(this.toggle, this));

if the buttons are placed outside of the calendar area 'click.daterangepicker' gets called. This will toggle the displaying of the calendar.  Basically, you have two events listening to the same thing. 


But keep in mind, the option alwaysDisplayCalendars is not respected with that happens causing the calendar to close.

dangrossman:
That screenshot doesn't look like it's of this library. The click event is triggered when you click on something. There are many things in this widget listening for clicks on them and doing different things in response. Only "this.element" listens for a click event then calls `toggle` as you quoted. Additionally, the `toggle` function's code is unrelated to what the `alwaysShowCalendars` option does, which has to do with whether the widget displays calendars next to the list of predefined ranges when the custom range option isn't currently selected. There aren't any predefined ranges in your screenshot, so the option wouldn't have any applicability here. This pull request makes no sense, your screenshot makes no sense, your comments make no sense.

PR #2079, opened 2020-04-06
Add option timePickerHours

Aurelie-feyzin:
Allows you to choose the times that appear in the timePicker.
Warning force timePicker24Hour = true

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -908,33 +908,33 @@
 
             html = '<select class="hourselect">';
 
-            if (this.timePickerHours.length === 0 && !this.timePicker24Hour) {
+            if (this.timePickerHours.length == 0 && !this.timePicker24Hour) {
                 this.timePickerHours = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12];
             }
-            if (this.timePickerHours.length === 0 && this.timePicker24Hour) {
+            if (this.timePickerHours.length == 0 && this.timePicker24Hour) {
                 this.timePickerHours = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23];
             }
 
-            this.timePickerHours.map(i => {
-                let i_in_24 = i;
+            this.timePickerHours.map(function (hour) {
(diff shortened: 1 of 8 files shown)


daterangepicker.js
```diff
@@ -900,29 +908,33 @@
...
+                let disabled = false;
                 if (minDate && time.minute(59).isBefore(minDate))
                     disabled = true;
                 if (maxDate && time.minute(0).isAfter(maxDate))
                     disabled = true;
 
-                if (i_in_24 == selected.hour() && !disabled) {
-                    html += '<option value="' + i + '" selected="selected">' + i + '</option>';
+                if (i_in_24 === selected.hour() && !disabled) {
+                    html += `<option value="${i}" selected="selected">${i}</option>`;
```
mRoca, line 931: Are you sure this lib supports template strings, const & let ? IMHO, you should only use jQuery 1.10 compatible syntax

website/index.html
```diff
@@ -579,6 +582,12 @@ <h1 style="margin-top: 30px"><a id="config" href="#config">Configuration Generat
                                 </label>
                               </div>
 
+                                <div class="form-group">
+                                    <label for="timePickerHours">timePickerHours (number between 0 and 23 separate by a comma)</label>
```
mRoca, line 586: typo : `separated`

example/browserify/index.html
```diff
@@ -95,6 +95,12 @@ <h1 style="margin: 0 0 20px 0">Configuration Builder</h1>
                 </label>
               </div>
 
+              <div class="form-group">
+                <label for="timePickerHours">timePickerHours (number between 0 and 23 separate by a comma)</label>
```
mRoca, line 99: typo: separated

daterangepicker.js
mRoca: Same for arrow functions

Aurelie-feyzin: modified

Wernfried:
I added it in my fork: https://github.com/Wernfried/daterangepicker

Parameter `timePickerStepSize` controls all units. For your purpose it would be for example `timePickerStepSize: Duration.fromObject({hours: 2})` 
to get a step size of 2 hours. Minutes and seconds will be hidden.

PR #2132, opened 2020-07-30
Update UMD template to work in an ES module context.

### LINKED ISSUE #2131: Update UMD template to support ES modules
Within an ES module, `this` is undefined, rather than `window`. This causes `import './daterangepicker.js'` to fail on line 29: https://github.com/dangrossman/daterangepicker/blob/6075b419d6072010accda52080322e1487c4cc7b/daterangepicker.js#L29

The UMD template you're using was updated in September 2017 to detect `window` instead: https://github.com/umdjs/umd/blob/47e725735b854dfd93c1ca6fcebf1e018bd52606/templates/returnExportsGlobal.js#L32

mwanji:
Cf. https://github.com/umdjs/umd/blob/47e725735b854dfd93c1ca6fcebf1e018bd52606/templates/returnExportsGlobal.js

Fixes #2131

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -28,7 +28,7 @@
         // Browser globals
         root.daterangepicker = factory(root.moment, root.jQuery);
     }
-}(this, function(moment, $) {
+}(typeof window !== 'undefined' ? window : this, function(moment, $) {
     var DateRangePicker = function(element, options, cb) {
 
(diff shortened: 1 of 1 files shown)


mwanji:
Hi @dangrossman ,

Would it be possible to publish a new version of the library to npm? :)

PR #2139, opened 2020-08-11
daterangepicker support add custom style & keep selected date

xianghongai:
1. Add custom styles by configuring `wrapperClass`.


xianghongai:
2. `keepSelected: true/false`, do not reset another date when reselecting a date

xianghongai:
keepSelected-false

keepSelected-true

PR #2152, opened 2020-09-18
Ignore invalid data from input

gorbeia:
When original data in the input is not valid moment.js creates a date object but it is invalid. DateRangePicker then shows the month view filled of NaNs. This change makes DateRangePicker to ignore the input data if it can´t be properly parsed.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -304,6 +304,12 @@
                     start = moment(val, this.locale.format);
                     end = moment(val, this.locale.format);
                 }
+                if (!start.isValid()) {
+                    start = null;
+                }
+                if (!end.isValid()) {
+                    end = null;
+                }
                 if (start !== null && end !== null) {
                     this.setStartDate(start);
                     this.setEndDate(end);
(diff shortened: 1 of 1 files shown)


ahmadsyamim (approved):

daterangepicker.js
```diff
@@ -304,6 +304,12 @@
                     start = moment(val, this.locale.format);
                     end = moment(val, this.locale.format);
                 }
+                if (start != null && !start.isValid()) {
```
kevinpapst, line 307:
```suggestion
                if (start !== null && !start.isValid()) {
```

daterangepicker.js
```diff
@@ -304,6 +304,12 @@
...
+                    start = null;
+                }
+                if (end != null && !end.isValid()) {
```
kevinpapst, line 310:
```suggestion
                if (end !== null && !end.isValid()) {
```

kevinpapst (changes requested):
I have suffered from that bug as well, so I like the solution 👍 
Just a minor suggestion to use type safe null safe check

PR #2173, opened 2021-01-13
Add minSpan option

onezerotrash:
If you set minSpan option then selected date range cannot be shorter.

### DIFF
--- a/daterangepicker.css
+++ b/daterangepicker.css
@@ -189,7 +189,11 @@
   color: #999;
 }
 
-.daterangepicker td.in-range {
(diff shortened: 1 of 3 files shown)


Joeycool993 (approved):

Joeycool993:
great

nathan-da-silva:
Seems to be a bug where if you click the last date 3 times, it un-restricts the subsequent dates defined by the minSpan

Wernfried:
I added this feature in my fork https://github.com/Wernfried/daterangepicker

You can set parameter `minSpan` and `maxSpan`

PR #2217, opened 2021-09-16
Add events on user date selection (but not applied)

cavasinf:
Add two event `selected.date.start` & `selected.date.end` when user click on date.
Note : if `autoApply`, there's no `selected.date.end` trigger


cavasinf:
This allows us to select the week from a single choosen date.

Animation

PR #2224, opened 2021-10-11
WIP: Add documentation for data attributes

deepakmahakale:
This adds documentation and example for using data attributes


deepakmahakale:
@dangrossman Should we add all the attributes and their hyphen case mapping in a table?

PR #2231, opened 2021-11-23
Add isInvalidTime to handle similar cases as isInvalidDate

bpofficial:
Added the `isInvalidTime` function to allow validating times etc. This was useful for a particular use case where I want to disallow the selection of a time that is within a specific range of another.


880831ian:
I need this feature, but no one seems to be maintaining this project right now ＱＱ

bpofficial:
It’d be good to get this merged

Wernfried:
I added it to my fork: https://github.com/Wernfried/daterangepicker

PR #2307, opened 2023-01-24
fix: vite moment error

akki-jat:
It fixed issue when used in vite it shows localData is not a function error.

Also includes fix from #2228

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -22,7 +22,8 @@
             jQuery = require('jquery');
             if (!jQuery.fn) jQuery.fn = {};
         }
-        var moment = (typeof window != 'undefined' && typeof window.moment != 'undefined') ? window.moment : require('moment');
+        var moment = (typeof window != 'undefined' && typeof window.moment != 'undefined') ? window.moment : require('moment/moment');
+        if (typeof moment !== 'function' && moment.hasOwnProperty('default')) moment = moment['default'];
         module.exports = factory(moment, jQuery);
     } else {
         // Browser globals


akki-jat:
@dangrossman Please take a look.

ghost:
Would be very happy to see this PR approved soon @dangrossman, it would fix our builds and prevent us from looking for alternatives...

dangrossman:
> Would be very happy to see this PR approved soon @dangrossman, it would fix our builds and prevent us from looking for alternatives...

You're waiting on a 11 year old library with no updates in years to fix your build process? Merge the PR in your copy, or find an alternative. I have no idea what Vite is and have no way of knowing if this PR will break others' builds. Someone put require('moment') in there on purpose, now it's being changed to require('moment/moment'). That must break something, or it'd always have been moment/moment. I know moment isn't changing APIs as they archived that project years ago.

Jogai:
At least set the project to archived then, to show visitors that its not under maintenance anymore.

MarcelYieldDD:
You have a point there @dangrossman. But it's just that I love the simplicity and integration with Bootstrap in this picker so a big #h5yr for that (and I so want to keep using this).

Now, JS is not my native language, but from what I've read [here](https://github.com/jalaali/moment-jalaali/pull/226) it shouldn't break anything, and sounds like a comparable issue:

> When requireing a package in a js file, it will point to the main field in package.json, which includes name of the main file of the package and its relative address to package.json file. But it is possible to be extra specific about the file you wish to include and add the file name after package name, it's a bit overuse when package exports only one file, but useful when a package has multiple files with different functionalities to offer.
So using require('moment/moment') is more detailed way of using require('moment'), it should not break older usages.

ghost:
You have a point there @dangrossman. But it's just that I love the simplicity and integration with Bootstrap in this picker so a big #h5yr for that (and I so want to keep using this).

Now, JS is not my native language, but from what I've read [here](https://github.com/jalaali/moment-jalaali/pull/226) it shouldn't break anything, and sounds like a comparable issue:

>When requireing a package in a js file, it will point to the main field in package.json, which includes name of the main file of the package and its relative address to package.json file. But it is possible to be extra specific about the file you wish to include and add the file name after package name, it's a bit overuse when package exports only one file, but useful when a package has multiple files with different functionalities to offer.
So using require('moment/moment') is more detailed way of using require('moment'), it should not break older usages.

shravyaRB:
I was able to fix the error by defining an alias
```
alias: {
      moment: path.resolve(__dirname, 'node_modules/moment/moment.js'),
}
```

clipsmm:
> I was able to fix the error by defining an alias
> 
> ```
> alias: {
>       moment: path.resolve(__dirname, 'node_modules/moment/moment.js'),
> }
> ```

Does this still work. Still getting the same error after this

PR #2327, opened 2023-06-20
minHour and maxHour

raponiwalter:


SBagaria2710 (approved):

Wernfried:
In my fork https://github.com/Wernfried/daterangepicker I added parameter `isInvalidTime` where you can set it.

```
isInvalidTime: (time, side, unit) => {   
   const minHour = 8;   
   const maxHour = 18;
   if (unit == 'hour') {
      return time.hour < minHour || time.hour > 18;
   } else {
      return false;
   }
}
```

PR #2346, opened 2023-12-18
added 'showTwoMonth' parameter

Eventpicker:
i had a problem specifically on mobile applications. the date picker with 2 months displayed went over the whole screen and was difficult to use. so i introduced a new parameter that only works with datepicker. 'showTwoMonth' is a boolean and is true by default. you can set it to false and only 1 month will be displayed.

if timepicker is activated this parameter has no effect

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -395,7 +395,7 @@
             }
         }
 
-        if(!this.showTwoMonth){
+        if(!this.showTwoMonth && !this.timePicker){
             this.container.find('.drp-calendar.right').hide();
         }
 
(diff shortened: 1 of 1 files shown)


mikeluyties:
This is a great addition. I also have this issue on mobile and want to limit it to a single calendar, but still with the date range functionality.

kadekdodikwirawan:
I was able to create a single calendar while still keeping the date range functionality by adding:  **linkedCalendars: false**

```js
new DateRangePicker(dateRangePicker, {
    startDate: new Date(),
    endDate: new Date(),
    parentEl: dateRangePicker.parentNode,
    linkedCalendars: false
});
```

And in the SCSS:

```scss
.daterangepicker {
    .drp-calendar.right {
        display: none !important;
    }
}
```

Eventpicker:
The problem with your solution is that you lose the “next” button.
The "next" button is displayed with the “showTwoMonth” parameter when this is set to true or false.

kadekdodikwirawan:
> The problem with your solution is that you lose the “next” button. The "next" button is displayed with the “showTwoMonth” parameter when this is set to true or false.

The next button is still apear, just make sure to set ``` linkedCalendars: false ```
image

Eventpicker:
Yes, that's true, sorry about that.

Your solution also works!
But I prefer a solution where I can continue using the other functions.
The feature `linkedCalendars: true` has been lost as a result.

PR #2349, opened 2024-01-15
feat: Year reload without changing the selection

youboll:


youboll:
Please do not accept it, i sent this by accident

PR #2360, opened 2024-08-09
add option timePickerMinutes

yk-eta:
add option timePickerMinutes to allow not to display minute in time picker

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -50,7 +50,7 @@
         this.timePicker = false;
         this.timePicker24Hour = false;
         this.timePickerIncrement = 1;
-        this.timePickerMinutes = false;
+        this.timePickerMinutes = true;
         this.timePickerSeconds = false;
         this.linkedCalendars = true;
         this.autoUpdateInput = true;
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -50,6 +50,7 @@
         this.timePicker = false;
         this.timePicker24Hour = false;
         this.timePickerIncrement = 1;
+        this.timePickerMinutes = false;
         this.timePickerSeconds = false;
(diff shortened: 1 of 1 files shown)


daterangepicker.js
```diff
@@ -50,6 +50,7 @@
         this.timePicker = false;
         this.timePicker24Hour = false;
         this.timePickerIncrement = 1;
+        this.timePickerMinutes = false;
```
Ekzosta: Should be `true` by default, as currently it's shown by default

Ekzosta (changes requested):

daterangepicker.js
yk-eta: ya, you are right

Wernfried:
I added this feature in my fork: https://github.com/Wernfried/daterangepicker

I created a generic parameter `timePickerStepSize`. Just set `timePickerStepSize` to a value greater or equal to  `Duration.fromObject({hours: 1})`

> Default: Duration.fromObject({minutes:1})
> Set the time picker step size.
> Must be a luxon.Duration or number of seconds or a string according to [ISO-8601](https://github.com/Wernfried/daterangepicker/blob/master/ISO-8601) duration.
> Valid values are 1,2,3,4,5,6,10,12,15,20,30 for Duration.fromObject({seconds: ...}) and Duration.fromObject({minutes: ...}) and 1,2,3,4,6,(8,12) for Duration.fromObject({hours: ...}).
> Duration must be greater than minSpan and smaller than maxSpan.
> For example timePickerStepSize: 600 will disable time picker seconds and time picker minutes are set to step size of 10 Minutes.
> Overwrites #timePickerIncrement and #timePickerSeconds

PR #2365, opened 2024-09-26
Added minSpan option

diogo24m:
Added a minSpan option

If end date does not equal or exceed start date + minSpan, use the start date + minSpan as end date.

As safety measure if minSpan exceeds maxSpan, use maxSpan value for minSpan.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -38,6 +38,7 @@
         this.endDate = moment().endOf('day');
         this.minDate = false;
         this.maxDate = false;
+        this.minSpan = false;
         this.maxSpan = false;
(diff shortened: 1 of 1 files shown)


pradyvrp:
I was searching for this feature and found it here but not available in a release version. Last release is also 4 years ago. It's kind of surprising how this feature wasn't included till now. Thanks for your code.

diogo24m:
I'm afraid this will never get merged

Wernfried:
I added this feature in my fork https://github.com/Wernfried/daterangepicker

You can set parameter `minSpan` and `maxSpan`

PR #2372, opened 2024-12-26
support for showSelectedDays option in custom range

kavvya97:
Fix for #2364

Added a new optional feature option `showSelectedDays` for Custom Range option. This feature is by default hidden.
The label for this feature is customizable through locale options. 

Note: I have disabled this feature when timepicker is enabled. this is to ensure that we don't want to show the days in fractional value. However, I can add this option if desired.

image

### DIFF
--- a/daterangepicker.css
+++ b/daterangepicker.css
@@ -92,7 +92,8 @@
   float: none;
 }
 
-.daterangepicker.single .drp-selected {
+.daterangepicker.single .drp-selected,
+.daterangepicker.single .drp-selected-dates {
   display: none;
 }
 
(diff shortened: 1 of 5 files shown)


Wernfried:
I added this feature in my fork at https://github.com/Wernfried/daterangepicker

You can define the format more granular with parameter `locale.durationFormat`

Format a custom label for selected duration, for example '5 Days, 12 Hours'.
Define the format either as string, see [Duration.toFormat - Format Tokens](https://moment.github.io/luxon/api-docs/index.html#durationtoformat) or an object according to [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#options), see [Duration.toHuamn](https://moment.github.io/luxon/api-docs/index.html#durationtohuman).

PR #2380, opened 2025-07-10
Add Month Range Picker Feature

hatimmakki:
This PR adds a month range picker feature to the daterangepicker library. 

## Changes:
- Added isMonthRange option to enable month-only selection
- Modified calendar rendering to show month/year selection when in month range mode
- Updated month navigation logic for month range picker
- Maintained backward compatibility with existing day picker functionality

## Usage:
```javascript
.daterangepicker({
    isMonthRange: true,
    locale: {
        format: 'MM/YYYY'
    }
});
```

This feature allows users to select month ranges instead of specific dates, which is useful for reporting, analytics, and other use cases where month-level granularity is preferred.


maricavor:
The problem that in month-only mode the plugin uses a different calendar structure and never sets `leftCalendar.calendar / rightCalendar.calendar`, but `hoverDate` still tried to read `calendar[row][col]`.
An early return was added for month mode so the hover logic is skipped. That’s fine for the month picker, since it’s single-date selection and doesn’t need range hover highlighting.
```
hoverDate: function(e) {
    //ignore dates that can't be selected
    if (!$(e.target).hasClass('available')) return;

    // Month range mode uses a different calendar structure; skip hover logic
    if (this.isMonthRange) return;

    ...
```

PR #2382, opened 2025-09-29
Feature: options.singleCalendar

eoghanmurray:
I noticed #1883 after implementing this.

Unlike that one, this PR maintains ability to have a start and end time selection.

It just renders the 'right' calendar instead of both; this is so that the end of the range is always in view (which made more sense for my usecase).

Also unlike #1883, this one handles `updateMonthsInView`.

### DIFF
--- a/daterangepicker.js
+++ b/daterangepicker.js
@@ -41,6 +41,7 @@
         this.maxSpan = false;
         this.autoApply = false;
         this.singleDatePicker = false;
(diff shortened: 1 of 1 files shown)


Wernfried:
I also added this feature in my fork https://github.com/Wernfried/daterangepicker

You can set it with parameter `singleMonthView: true`

